PR #32, opened 2013-10-31
Add facility to preserve go installation

tmc:
Some go applications rely on having the go installation (GOPATH and/or GOROOT) to be around at runtime. This adds a check for a ".gokeep" file to enable such.

### DIFF
--- a/bin/compile
+++ b/bin/compile
@@ -79,10 +79,23 @@ else
     echo " done"
 fi
 
-GOROOT=$cache/$ver/go export GOROOT
-GOPATH=$build/.heroku/g export GOPATH
-PATH=$GOROOT/bin:$PATH
+if test -d $build/.goroot
+then
+    echo -n "-----> Preserving GOROOT due to .goroot directory presence"
+    cp -R $cache/$ver/go/* $build/.goroot
+    GOROOT=$build/.goroot export GOROOT
+else
+    GOROOT=$cache/$ver/go export GOROOT
+fi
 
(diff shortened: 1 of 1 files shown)


freeformz:
What is the use case here?

tmc:
I was running a tool that expected the go source to be on disk (pythia in particular).

freeformz:
@tmc So you were using pythia as your `web` Procfile entry to view the source code of your app ?

freeformz:
Sometimes I wonder if the buildpack should set `GOPATH=/app` and then set the pushed code up properly inside of that (`/app/src/<ImportPath>/`). This way using pythia is at worst a matter of vendoring a copy into your repo and at best just supported (somehow, not sure how) out of the box.

I'm not exactly -1, but also not exactly +1 on keeping the go toolchain (basically GOROOT) around. I'm mostly just wondering how many people would find that useful vs. the amount of bloat it would add (even optionally) to the slug.

freeformz:
@tmc Even if `GOPATH=/app/src/<ImportPath>` and `GOROOT=.goroot` (or some such) how are you planning to run pythia / ssaview exactly?

freeformz:
I'm going to close this implementation wise. I have some thoughts about supporting this better including actually setting up a real GOPATH (see the docker branch), which I'll get to in August.

tmc:
@freeformz I'm not sure what your question is there on the 30th -- I deploy them with the proposed patch (or a version of it)

PR #57, opened 2014-09-17
Work around incorrect $GIT_DIR on Heroku

attilaolah:
It appears that `$GIT_DIR="."` in the deploy environment, while we're not running in `/app/tmp/repo.git`, so any call to `git` fails.

This workaround hard-codes the absolute path, and frankly it seems like a bad idea but I could not come up with anything better, and our deploys are failing so this is kind of urgent.

@kr if you happen to know what happened to `GIT_DIR`, that could improve on this maybe.

### DIFF
--- a/bin/compile
+++ b/bin/compile
@@ -145,7 +145,7 @@ fi
 FLAGS=(-tags heroku)
 if test -n "$GO_GIT_DESCRIBE_SYMBOL"
 then
-    git_describe=$(git describe --tags --always)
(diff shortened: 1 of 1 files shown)


kr:
Well, I finally went and had to remove this feature entirely in 36476a80efb196676091c07acfe54e4443f974e3. (Not just because of the incorrect output, but also because Heroku is in the process of removing the git repo entirely from the build process.) Sorry!

PR #66, opened 2014-11-13
Fetch tags before running git describe

ryanbrainard:
`git describe --tags --always` works much better if the tags are up to date, so try to run `git fetch --tags` first.


kr:
Ah good call. Thanks!

Unfortunately, I had to ax this feature entirely. 36476a80efb196676091c07acfe54e4443f974e3

PR #68, opened 2014-12-09
Behave better when GO_GIT_DESCRIBE_SYMBOL is unset

jchesterpivotal:
Necessary when working in an environment using `set -u`. Eg:
- a Cloud Foundry buildpack
- a multi-buildpack setting -u

### DIFF
--- a/bin/compile
+++ b/bin/compile
@@ -143,7 +143,7 @@ then
 fi
 
 FLAGS=(-tags heroku)
-if test -n "$GO_GIT_DESCRIBE_SYMBOL"
+if [ -z GO_GIT_DESCRIBE_SYMBOL ] && [ test -n "$GO_GIT_DESCRIBE_SYMBOL" ]
 then
     git_describe=$(git describe --tags --always)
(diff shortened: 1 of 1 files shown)


kr:
Unfortunately, I've had to remove this feature entirely in 36476a80efb196676091c07acfe54e4443f974e3. Here's the description from that commit, for easy reference:

>This feature has never worked correctly.

>See https://gist.github.com/kr/f22beb85af1ca27890f3
for some analysis. The buildpack runs inside a
pre-commit hook, so the to-be-committed object isn't yet
present in the history to be queried by git commands.
This means every push will see the previous head as its
"current" commit, and the very first push will fail
entirely (since there is no history at all yet).

>Further, Heroku is in the process of removing all git
artifacts from the build process. This is meant to
accomodate other non-git deployment systems and make the
build process consistent for all deployment transport
mechanisms.

>Until Heroku provides an explicit means of getting some
identifying information into the build process, it seems
there is not much we can do on this front.

PR #78, opened 2015-03-20
Add a way to run commands after compilation

attilaolah:
This is useful e.g. for running migrations, when the tools required are not going to be built (e.g. because they are vendored in `Godeps`).

`POST_BUILD_COMMAND` can be something simple, or it can point to a script inside the repo.

Here's an example post-install script:

``` sh
export PATH="$PATH:$(godep path)/bin"
godep go install github.com/some/vendored/cmd/migrate
migrate -env=heroku
```

---

I am of course open to suggestions, if there's a better way to do this, however this seemed to be the simplest solution.

### DIFF
--- a/bin/compile
+++ b/bin/compile
@@ -153,6 +153,12 @@ else
     go get -tags heroku ./...
 fi
 
+if test -f "$env_dir/POST_BUILD_COMMAND"
+then
+    echo "-----> Running: $(cat $env_dir/POST_BUILD_COMMAND)"
+    source "$env_dir/POST_BUILD_COMMAND"
+fi
+
 rm -rf $build/.heroku
 
 mkdir -p $build/.profile.d
--- a/bin/compile
+++ b/bin/compile
@@ -155,8 +155,12 @@ fi
 
 if test -f "$env_dir/POST_BUILD_COMMAND"
 then
-    echo "-----> Running: $(cat $env_dir/POST_BUILD_COMMAND)"
-    source "$env_dir/POST_BUILD_COMMAND"
+    for key in "$env_dir"/*
+    do
+        ENV="$ENV $(basename $key)=\"$(cat $key)\""
+    done
+    echo "-----> Running: $(cat "$env_dir"/POST_BUILD_COMMAND)"
+    eval "$ENV $(cat "$env_dir"/POST_BUILD_COMMAND)"
 fi
 
 rm -rf $build/.heroku


freeformz:
I'd rather not tie this to an environment variable. My thoughts are either:

``` bash
if make --dry-run heroku-post-compile
then
  make heroku-post-compile
fi
```

or

``` bash
if test -f bin/post_compile
then
  chmod +x bin/post_compile
  bin/post_compile
fi
```

The former seems more "Go like". The later is in line with the python build pack.

ojacobson:
I'm broadly in line with the latter, changing it as

``` bash
if test -x bin/post_compile
then
  bin/post_compile
fi
```

i.e., respect +x permissions as set by the author, rather than forcing them. However, that probably plays interestingly with Windows users, who may not have a great way of setting permissions on Git-tracked scripts. It's probably worth asking @kennethreitz what his original rationale for chmod vs trusting permissions was.

`make --dry-run` generates a bunch of output, and the `make` interface for task goals (rather than file goals) is error-prone, so I'm not a huge fan of supporting it or encouraging it.

attilaolah:
I'm also :-1: for enforcing `make` on this, some folks might use other build tools.

I'm also not a fan of the env var, but I couldn't really come up with anything better. I'd just go with @ojacobson's suggestion, now that he mentions it, and not worry about Windoze users. They can probably set the permissions with the `git` command anyway, even if that is not reflected locally.

kr:
Note that git hooks (pre-commit, post-update, etc) use hyphens, not underscores. Maybe consider doing the same here.

freeformz:
This is something very tricky to do right in the context of a build.

I think we need to find some patterns that help us run the migrations explicitly in the context of an app and it's config instead of during build time, which is increasingly being divorced from the git push model (see the new docker stuff, pipelines, etc).

I plan to use the migrate package (https://github.com/mattes/migrate) to explore this though.

Thanks for your input, but I'm closing this for now.

frederikhors:
Any news on this after years?

I just need some command (pre/post compile) which generates/builds my static assets.

Is there a way to do this considering the dynos's ephemeral disk?

freeformz:
@frederikhors What tool is being used to generate / build your static assets? If it's webpack or another tool in the node ecosystem, then you should use the heroku node js buildpack to build them after/before the go buildpack via: `$ heroku buildpacks:add heroku/nodejs`. See more details [here](https://devcenter.heroku.com/articles/using-multiple-buildpacks-for-an-app#adding-a-buildpack).

frederikhors:
@freeformz thanks.

I just use some Go tools. So I need a way to do something with go during build. **How to?**

https://stackoverflow.com/questions/54410715/heroku-golang-asset-compilation-should-happen-during-builds-but-how

freeformz:
My opinion is that unless the thing you are running is something very specific and common to Go ecosystem, then this should be done in a different buildpack, not the Go buildpack.

For instance I am happy to entertain the idea of adding something like executing a [mage file](https://github.com/magefile/mage) or some other go tooling that is likely to be common.

I'm just against the idea of adding an arbitrary hook.

So, what tools are you using and how are you using them?

If they are home grown tools then my recommendation would be to us the [inline buildpack](https://github.com/kr/heroku-buildpack-inline) to execute a code local "./bin/compile" script to do what you need to do.

If you are interested in the idea of using Mage to drive that, I can explore implementing it on a branch and would love your feedback / testing of that branch.

Also open to other ideas.

jmorrell:
@freeformz I'd ask you to revisit that decision. We can't predict every use case a golang customer will need and having `pre` and `post` hooks gives the user a ton of flexibility to solve their own problems without having to understand the intricacies of the buildpack system. Users like using Heroku because they don't have to think about things like that.

If they can google "go build script heroku" and add one line to an existing file, that's way better than having to understand what a buildpack is, why they would want the "inline buildpack", how to order that alongside the go buildpack, etc.

cc @sowjumn @CaseyFaist

freeformz:
@jmorrell So the nodejs buildpack has a method to run something arbitrary before and/or after it does stuff?

jmorrell:
@freeformz Yes: https://devcenter.heroku.com/articles/nodejs-support#heroku-specific-build-steps

(though we're in the process of supporting the community default that has emerged in the past couple of years https://github.com/heroku/heroku-buildpack-nodejs/issues/583)

We save these scripts in dep-tracker and can pull them to see what people are doing with that flexibility if you're interested. Many times it's working around things the buildpack should add, but often it's things that are specific to that app or org.

There are other ways to do that: `heroku.yml`, inline buildpack, custom buildpack, but they are all heavier and place more burden on the user

freeformz:
Like I said, I'm down to doing it via something like mage. My idea would be that if `heroku:pre-install` / `heroku:post-install` targets exist they will be executed. I can see some possible uses for other hooks as well. Maybe. 

^ ends up becoming a fancy set of hooks and pushes things in a direction where the buildpack isn't executing an arbitrary user supplied bash blob.

Thoughts? Are there other tools currently in the go ecosystem that others would prefer to use? Or do people really really really want a "bash command here" thing?

jmorrell:
I like supporting tools in the ecosystem, but unless mage is ubiquitous then we're forcing tools onto the user. If we want to take the position "All go projects should use mage" then I'm down with that, but I suspect that's not the case here.

We are using comments in `go.mod` to specify runtime version: https://github.com/heroku/heroku-buildpack-go#go-module-specifics. Perhaps something similar could be used for "run this command"? 

```
// +heroku buildCommand scripts/build.sh
```

(Though aesthetically I don't really like this. It does solve the problem well and would be super easy for customers)

Perhaps @sowjumn could weigh in here given that she will be leading up a squad focused on this buildpack soon

freeformz:
I am tempted to say "All go projects should encode their misc tasks in a magefile", but I'm not really sure how accepted mage is atm.

frederikhors:
For what my opinion is worth:

- I do not like the "magefile" solution and I agree with @jmorrell that we should not force the user to use specific tools
- I really like the idea of using go.mod as @jmorrell explains (elegance comes later if I can do my job)
- heroku.yml could be another way
- python users have `post_compile` hook too: https://www.codementor.io/samueljames/a-workaround-heroku-s-ephemeral-file-system-e6w341zqa

Issue #63, opened 2014-10-30
Use canonical import path if present


kr:
See http://tip.golang.org/doc/go1.4#canonicalimports

The canonical import path, if present in a package being built, should be used in preference to our existing sources of import path (`Godeps.json`, `.godir`).

freeformz:
@kr I would assume that godep (or similar tooling) would DTRT when vendoring here instead of pushing this down to the build pack.

kr:
Yes, godep should do the right thing. But this issue also has to do with projects that don't use godep. It used to be that, for such projects, the buildpack had no way to know the import path (which is why we defined the `.godir` convention). Now, the buildpack can check for a canonical import path magic comment.

Of the three possible ways to figure out the app's import path (`Godeps.json`, `.godir`, canonical import path), the canonical import path is the only one defined and approved by the Go team and used by stock go tooling that ships with the compiler. So I think if the canonical import path is present, it should be used.

cyx:
Do you mean defining GOPATH as an environment variable and having the
buildpack read that?

kr:
> Do you mean defining GOPATH as an environment variable and having the
> buildpack read that?

No, please see the link above for the definition of a canonical import path. This is about determining the base import path for the app. It is orthogonal to the choice of GOPATH.

freeformz:
@kr My plan (pending some discussion and fact finding) is to remove support for non godep deploys, possibly adding in support for tools like https://github.com/kardianos/vendor and https://github.com/constabulary/gb as they become more feature complete.

elithrar:
@freeformz Before I consider hacking on anything, have you started any work on making `gb build` (and/or also `gb-vendor`) an option for the buildpack?

freeformz:
@elithrar Not as of yet as we've been focused on prepping some other stuff.

I'm happy to review your work when it's ready!

kr:
Some more facts for your finding: this thread was revived today. https://groups.google.com/d/msg/golang-dev/74zjMON9glU/4lWCRDCRZg0J

If we're lucky, all the vendoring tools (and users thereof) will converge on this convention, and the buildpack will have only to run `go install -vendor ./...` and care not which vendoring tool was used. :grin: 

(Of course, that doesn't solve the issue of how to find the import path of the repo root, but this issue provides a pretty clear way to handle that: canonical import paths.)

elithrar:
[...]

PR #83, opened 2015-06-16
Use the canonical import path if defined

freeformz:
This allows us to use a go standard to define the path for the code
instead of relying on `.godir`.

/cc #63

### DIFF
--- a/bin/compile
+++ b/bin/compile
@@ -35,6 +35,8 @@ urlfor() {
     esac
 }
 
+DEFAULT_GOVERSION=1.4.2
+
 mkdir -p "$1" "$2"
 build=$(cd "$1/" && pwd)
 cache=$(cd "$2/" && pwd)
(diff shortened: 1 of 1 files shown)


bin/compile
```diff
@@ -65,12 +82,21 @@ then
     ver=$(<$build/Godeps/Godeps.json jq -r .GoVersion)
 elif test -f $build/.godir
 then
+    warn "WARNING: .godir is deprecated and will be removed."
+    warn "Please use godeps or a canonical import path. For instructions:"
+    echo_urls
     name=$(cat $build/.godir)
-    ver=go${GOVERSION:-1.4.2}
+    ver=go${GOVERSION:-$DEFAULT_GOVERSION}
+elif go list -f {{.ImportComment}}
```
kr: I'm not sure this test does what you want. The exit status of this 'go list' command can be 0 even when there's no import comment. (And if there were an import comment, wouldn't this print out the value to the user?)

Maybe `elif test ! -z "$(go list -f {{.ImportComment}})"`?

freeformz: Fixed. I dunno what I was thinking.

kr:
LGTM
FWIW
:)

freeformz:
Closing this in favor of #88

PR #85, opened 2015-06-30
Defaults for GOMAXPROCS

freeformz:
/cc @cyx

### DIFF
--- a/bin/compile
+++ b/bin/compile
@@ -168,3 +168,4 @@ rm -rf $build/.heroku
 
 mkdir -p $build/.profile.d
 echo 'PATH=$PATH:$HOME/bin' > $build/.profile.d/go.sh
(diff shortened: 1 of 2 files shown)


cyx:
LGTM. It seems a bit conservative, but I'm not sure about good tuning defaults for it either.

kr:
Note this is likely to change in Go 1.5 http://golang.org/s/go15gomaxprocs

freeformz:
@kr it will, but I don't think you'd actually want to set GOMAXPROCS to the cpu count for 1x / 2x dynos and the other types are == to the cpu counts of the underlying systems.

freeformz:
I will also likely modulate this based on actual tests and feedback at some point in the future, but this is likely better than the default of 1 (or the new 1.5 defaults) on Heroku.

PR #88, opened 2015-07-14
Use cannonical import path if it exists

freeformz:
This required more work than initially expected because go needs to be installed first.

Closes: #83, #63

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ### Unreleased
 
 * Use a canonical import path if it exists.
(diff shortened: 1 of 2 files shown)


freeformz:
As pointed out on Slack we'll also need to pull the customer's desired GO_VERSION otherwise the outcome won't be what people expect.

freeformz:
I'm going to need to sit on this for a while longer as I'm about to go on vacation and didn't get to it before hand.

freeformz:
I'm not sure this is a good idea and it's a huge conflict now, so closing. Please re-open if you want/need this and we can discuss.

PR #92, opened 2015-08-21
drop support for ancient versions of go

kr:
Go 1.2 was released in 2013 (and Go 1.2.1 in March of
2014). It is supplanted by three major versions more
recent—the current release is Go 1.5. There is no reason
for anyone to still be using these very old releases.

### DIFF
--- a/bin/compile
+++ b/bin/compile
@@ -3,35 +3,15 @@
 
 set -eo pipefail
 
-# Go releases for Darwin beginning with 1.2rc1
-# have included more than one build, depending
+# Go releases for Darwin
+# include more than one build, depending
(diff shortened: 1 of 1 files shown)


freeformz:
Thanks @kr!

I don't know if I can "just" drop this. I need to check some numbers wrt what apps are deployed using and how recently they pushed.

I may need to deprecate it for a while first as well, I need to double check with our support policy.

freeformz:
I took a look into some numbers late last week and I can't just drop these. 

But I am going to deprecate versions of Go < 1.3.3 (and maybe minor, not latest versions of Go < 1.5 i.e 1.4 + 1.4.1) and start displaying a warning message.

PR #94, opened 2015-08-26
Added vendoring experiment flag

sudhirj:


freeformz:
I can't do it this way w/o affecting users who don't want to use `vendor/`. I'll add support for the env var, so users can set it on their app and then it will be used.

freeformz:
I added support in fd88e78fd1c29f3230858c9c5a311bb8c054d0d7

sudhirj:
:+1:

Issue #91, opened 2015-08-20
Support for Nut


ebramanti:
It would be great if this buildpack supported [Nut](https://github.com/jingweno/nut), a pretty popular dependency manager for Go. I'm working on this on a fork, but I'm curious if anyone else is interested in having this added to the buildpack.

bithavoc:
:+1:

ebramanti:
@bithavoc I've gotten it to work somewhat, but I've pretty much broken the buildpack's normal functionality to do it. Can you try forking and seeing if you can make any progress?

freeformz:
Personally (not speaking for my position @ Heroku) I'd much rather spend time adding support for vendor-spec, which all of these tools should support, especially as Go 1.6 approaches. I am working towards that with godep atm (albeit slowly).

With that said (putting my Heroku hat on). Is there sufficient interest in supporting NUT? The author of which also works @ Heroku. If you want support for https://github.com/jingweno/nut please +1 this issue.

ebramanti:
:+1:

ebramanti:
@jingweno would you be interested in seeing support for this with Heroku?

owenthereal:
I :+1: to this. But that may require changes to `nut` to declare Go version etc. that the buildpack needs.

bithavoc:
I second @freeformz on this.

ebramanti:
@bithavoc second on vendor support or Nut?

freeformz:
@jadengore IMO Nut + godep + buildpack supporting the vendor spec is the best outcome, but I understand the argument for interim support of Nut though, as long as nut provides all of the information necessary to compile the code reproducibly (like go version).

ebramanti:
@freeformz I have Nut working specifically for the app I'm working on, but my fork has significantly diverged and I don't have too much shell scripting experience to know if this is ready for public release.

What I plan to do is iterate on my work and have a PR for discussion/amendments so that maybe we can get this merged in with Nut support.

freeformz:
@jadengore SGTM Having something to review is probably best.

freeformz:
Can Nut not support the vendor-spec instead?

freeformz:
We're not going to be supporting nut, in fact I'd like to reduce the number of supported tools. Hopefully dep and the other plans around future dependency management stuff in go 1.10/11/etc will make that a reality.

PR #97, opened 2015-09-26
Add Nut support to buildpack

ebramanti:
As discussed in #91, I've made a pull request with Nut support for the Go Heroku buildpack. 

Implementation details bulleted below:
- Add Nut.toml validation step and binaries for TOML validation/decoding
- Add Nut install step and Nut binary to buildpack

I've also made a couple of assumptions for this to work properly. I need some input from @jingweno on his thoughts for the assumptions below, since we will need to update the Nut repo to reflect this if so.
1. In `Nut.toml`, I now use the `import_path` string under the `application` table to store the import path that the buildpack needs to access your repository on Heroku. Here is an example of the new structure below, from Nut's repo example:
   
   ``` toml
   [application]
   
   name = "hello_world"
   import_path = "github.com/jingweno/hello-world" # the new convention
   version = "0.0.1"
   authors = ["Your Name <you@example.com>"]
   ```
2. For now, I am not specifying the version of Go in the `Nut.toml` file. It would be doable in the application table, but maybe some discussion is warranted first before I add this as well. I already feel like I am making a big assumption with 1.

@bithavoc, I welcome input as well on this PR since you mentioned you wanted support in #91. @freeformz, let me know if I need to make any changes to this PR so we can get this merged! 

And finally, big thanks to @jingweno for making Nut. It's my favorite Go dependency management tool! :smile:


freeformz:
Sorry I was out on vacation for a bit.....

Somethings to note while I consider this:
1. We need instructions for building/re-building/updating the toml tooling.
1. Can nut not support https://github.com/kardianos/vendor-spec instead of having it's own config file? I'm much more interested in moving that way, even for godep.
1. please dry up the version selection so multiple paths use the same default version.

ebramanti:
@freeformz I apologize for taking so long to get back to you about this branch. Here are my responses:
1. For building/updating TOML tooling, we go through the same process as `godep` and `jq`, which are also undocumented. Would you like me to just add a broad overview of how to build all of our linux package dependencies?
2. We need to ping @jingweno on this point because currently Nut does not conform to `vendor-spec`. This would remove the TOML dependencies, but I think this would be a big change for Nut. 
3. In the meantime, I could add a version number for go, such as `GoVersion`, to the TOML file. This would allow users to version their particular version of Go instead of the current 1.5 version I have them locked into.

freeformz:
Sorry it's taken so long to respond as well...

This is something we (Heroku) would have to support beyond just being able to build code via the buildpack. ATM we're not ready to support it.

I'd encourage Heroku users to open a support ticket asking for this if they want it.

If there is enough interest we'll certainly consider supporting it.

@jingweno is interested in contributing back to the bp sometime in the future though when he feels that nut is ready. 

In the meantime I'm going to close this issue.

ebramanti:
@freeformz Agree with your decision :+1:, it seems that Nut has not seen a lot of contribution or love as of late. Thanks for responding with feedback, and for helping me learn more about Heroku's buildpack system.

PR #109, opened 2016-02-02
Only use env_dir if provided.

lstoll:
Otherwise, we'll error out trying to create it. It seems to be optional, judging by the ruby buildpack. So we can just skip creating it, and let the existence check later fail.

/cc @fabiokung @freeformz

### DIFF
--- a/bin/compile
+++ b/bin/compile
@@ -105,6 +105,7 @@ GO_LINKER_VALUE=${SOURCE_VERSION}
 # allow apps to specify cgo flags and set up /app symlink so things like CGO_CFLAGS=-I/app/... work
 if [ ! -z "$env_dir" ]
 then
+    mkdir -p "$env_dir"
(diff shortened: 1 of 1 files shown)


bin/compile
```diff
@@ -103,8 +103,9 @@ finished() {
 GO_LINKER_VALUE=${SOURCE_VERSION}
 
 # allow apps to specify cgo flags and set up /app symlink so things like CGO_CFLAGS=-I/app/... work
-if [ -d "$env_dir" ]
+if [ ! -z "$env_dir" ]
 then
+    env_dir=$(cd "$env_dir/" && pwd)
```
freeformz, line 109: we should mkdir -p here first.

freeformz:
good catch, see note.

freeformz:
Thanks!

PR #110, opened 2016-02-09
Let's talk about plain way

dz0ny:
So instead of relying on new tools, we can support both vendoring via vendor experiment and locking packages by committing them to repository. All the user needs to setup is $GO_APP which in most cases translates to users app, but this can be automated if there is desire...

One example: https://github.com/dz0ny/arso-api

### DIFF
--- a/bin/compile
+++ b/bin/compile
@@ -104,7 +104,7 @@ then
     mkdir -p "$env_dir"
     env_dir=$(cd "$env_dir/" && pwd)
     ln -sfn $build /app/code
-  for key in CGO_CFLAGS CGO_CPPFLAGS CGO_CXXFLAGS CGO_LDFLAGS GO_LINKER_SYMBOL GO_LINKER_VALUE GO15VENDOREXPERIMENT GOVERSION
+  for key in CGO_CFLAGS CGO_CPPFLAGS CGO_CXXFLAGS CGO_LDFLAGS GO_LINKER_SYMBOL GO_LINKER_VALUE GO15VENDOREXPERIMENT GOVERSION GO_APP
     do
         if [ -f "$env_dir/$key" ]
         then
@@ -124,6 +124,10 @@ then
     name=$(<$build/Godeps/Godeps.json jq -r .ImportPath)
     ver=$(<$build/Godeps/Godeps.json jq -r .GoVersion)
     TOOL="godep"
+elif (test -d "$build/src" && test -n "$(find "$build/src" -type f -name '*.go' | sed 1q)" && test $GO_APP)
+then
+    ver=${GOVERSION:-$DEFAULT_GO_VERSION}
+    TOOL="native"
 elif (test -d "$build/src" && test -n "$(find "$build/src" -type f -name '*.go' | sed 1q)")
 then
     ver=${GOVERSION:-$DEFAULT_GO_VERSION}
@@ -186,7 +190,7 @@ case $TOOL in
         GOBIN=$build/bin export GOBIN
         GOPATH=$build/.heroku/go export GOPATH
         PATH=$GOROOT/bin:$PATH
-        
+
         p=$GOPATH/src/$name
         mkdir -p $p
         cp -R $build/* $p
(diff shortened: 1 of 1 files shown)


freeformz:
AFACT this really only works if the Go community decides that the right way to manage projects is to have a $GOPATH per project. I don't see that happening atm or am I missing something?

PS: At some point I may no longer be able to rely on environment variables from the app either, requiring that we do something else for env vars as well.

nathany:
> Go 1.5 introduced experimental support for vendoring, enabled by setting the GO15VENDOREXPERIMENT environment variable to 1. Go 1.6 keeps the vendoring support, no longer considered experimental, and enables it by default. http://tip.golang.org/doc/go1.6

All tools that use `go build` should be adopting the use of the `vendor/` folder so that which tool is used to copy the code doesn't matter and this build pack won't need specific support for any particular tool. The one possible exception is `gb` which is a completely separate build toolchain.

freeformz:
@nathany yes, agreed, but this PR isn't for that AFAICT.

nathany:
Precisely. There should be no need for $GO_APP.

freeformz:
I realized the other day that I can look at the pushed code and eliminate any imports that are in vendor and in $GOROOT. Any remaining imports are self referential and define the root of the project in GOPATH. If there are none, then I can just stick the app anywhere in a temp $GOPATH and compile it. This is going to happen, just need to do the footwork.

But that doesn't cover all uses and env variables in build packs _may_ be going away, so I was also thinking about a "better" way to have customers relay information like go version, ago flags, custom install package paths (./cmd/... + github.com/tool/migrate/... vs ./...), etc. We can either use the existing app.json file or another idea I had was a special `heroku` package in the root of Go project that uses a small DSL package to declare these things, then the bp builds that package and uses it to get the data out. thoughts?

davidjahn:
Hi @freeformz , we have solved this problem in the Cloud Foundry go-buildpack, which is forked from here:

https://github.com/cloudfoundry/go-buildpack/blob/develop/bin/compile#L162-173
https://github.com/cloudfoundry/go-buildpack/blob/develop/bin/compile#L290-L327

Our solution is to just have the user define a `GOPACKAGENAME` env variable which is the package name of your app.

Just wanted to say that we are in favor of merging this PR or something like it to support native vendoring in Go. We added the functionality because people were asking for it, but it would be nice if it were upstream in Heroku so we don't diverge too much. Thanks!

freeformz:
@davidjahn environment variables aren't passed in to the detect phase, so this won't work on Heroku.

jvshahid:
@freeformz so where is the `GOVERSION` on [this line](https://github.com/heroku/heroku-buildpack-go/blob/20a3e7129e24db6bbb269ecdea69015eb53b97e4/bin/compile#L176) coming from ? My understanding is that the environment variables are propagated using `env_dir`

freeformz:
@jvshahid They come from [here](https://github.com/heroku/heroku-buildpack-go/blob/20a3e7129e24db6bbb269ecdea69015eb53b97e4/bin/compile#L149-L160). `$env_dir` is set [here](https://github.com/heroku/heroku-buildpack-go/blob/20a3e7129e24db6bbb269ecdea69015eb53b97e4/bin/compile#L110). There is no equivalent for [detect](https://devcenter.heroku.com/articles/buildpack-api#bin-detect-usage) though.

freeformz:
@jvshahid Specifically what I've found is that, on Heroku, [on this line](https://github.com/cloudfoundry/go-buildpack/blob/94448874d2e81acee92c79cde1c86c750f344a3c/bin/detect#L12) `$GOPACKAGENAME` will always be zero because there is no `$env_dir` passed in and it's contents aren't sourced in detect.

PS: Also, depending on the detection order, [this line](https://github.com/cloudfoundry/go-buildpack/blob/94448874d2e81acee92c79cde1c86c750f344a3c/bin/detect#L11) can falsely pick up non go projects as a `vendor/` directory is also used in ruby and some ruby gems ship with .go file (I don't remember the specifics ATM, but I've seen this happen). I am not sure about other languages.

dz0ny:
I have setup workflow using bash buildpack that essentially moves project to root where it can be picked by go buildpack.

You can see this in effect here https://github.com/ubuntu-si/arso-api/blob/master/.heroku/run.sh

freeformz:
@dz0ny there are some other alternatives to that:

https://github.com/heroku/heroku-buildpack-multi-procfile
https://github.com/lstoll/heroku-buildpack-monorepo

Both of these are used internally.

PR #113, opened 2016-03-10
(WIP) GO15VENDOREXPERIMENT without godeps

xentek:
After growing frustrated with `godeps`, I decided to embrace the official
vendoring solution provided by `GO15VENDOREXPERIMENT`, which is now on
by default, and will come standard in 1.7 (won't be able to disable it).

This patch is my attempt to support `GO15VENDOREXPERIMENT` in a
dependency manager agnostic way.

Personally, I am using [`glide`](https://glide.sh) to manage my `vendor/` directory,
but realized that at this juncture this buildpack doesn't need to
support any particular tool to support a more generic vendored
dependency approach, since any front-end tool is irrelevant and not used
by go when the app is compiled.

This buildpack will now detect if `$GO15VENDOREXPERIMENT` is set to
`"1"`, and that a `vendor/` directory exists. If these are found, a
simple `go install` is run (using the appropriate flags). This should
work for those without any vendored dependencies, provided they have a
placeholder file in `vendor/` to ensure the directory can be pushed to
git.

This is my first attempt at hacking on a buildpack, so any feedback is
welcome and appreciated.

A couple of loose ends that may need attention:
- I don't test the version of go that is being used, to ensure it's 1.5 or greater.
- The text I `warn` users about in the build step should probably link
  to articles in your devcenter, recommend that you use `godeps`, etc.

### DIFF
--- a/bin/compile
+++ b/bin/compile
@@ -284,8 +284,6 @@ case $TOOL in
         mkdir -p $p
         cp -R $build/* $p
 
-        pkgs=$(<$build/Godeps/Godeps.json jq -r 'if .Packages then .Packages | join(" ") else "." end')
-
         case $ver in
         go1.5*)
             if test "$GO15VENDOREXPERIMENT" = "1"
--- a/bin/compile
+++ b/bin/compile
@@ -284,7 +284,7 @@ case $TOOL in
         mkdir -p $p
         cp -R $build/* $p
 
-        $pkgs=$(grep -l -R --exclude-dir=vendor --exclude-dir=.git "func main" $build | tr '\n' ' ')
+        pkgs=$(grep -l -R --exclude-dir=vendor --exclude-dir=.git "func main" $build | tr '\n' ' ')
 
         case $ver in
         go1.5*)
--- a/bin/compile
+++ b/bin/compile
@@ -284,6 +284,8 @@ case $TOOL in
         mkdir -p $p
         cp -R $build/* $p
         
+        warn $name
+        warn $p
         pkgs=$(grep -l -R --exclude-dir=vendor --exclude-dir=.git "func main" $p | tr '\n' ' ')
 
         case $ver in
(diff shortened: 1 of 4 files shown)


xentek:
As I was debugging this, I realized that the build environment  and what go expects are a bit at odds. This is why having something like Godeps.json handy is nice, as it can give you info about how to move the pushed code into the GOPATH properly before you install. I also discovered the elements marketplace, and am going to take a step back and create a slimmed down build pack that just supports glide directly (something I don't think is necessarily appropriate for the official build pack).

Closing this.

freeformz:
I would accept a PR that solves all of these problems and supports glide and/or govendor. They are next on my list of things to support.

bin/compile
```diff
@@ -128,6 +128,10 @@ elif (test -d "$build/src" && test -n "$(find "$build/src" -type f -name '*.go'
 then
     ver=${GOVERSION:-$DEFAULT_GO_VERSION}
     TOOL="gb"
+elif (test "$GO15VENDOREXPERIMENT" = "1" && test -d "$build/vendor")
```
freeformz, line 131: What about go1.6

bin/compile
```diff
@@ -268,10 +272,69 @@ case $TOOL in
         for f in bin/*-heroku; do mv "$f" "${f/-heroku}"; done
         rm -rf pkg
     ;;
+  govendor)
+        warn ""
+        warn "GO15VENDOREXPERIMENT, without another build tool, is unsupported."
```
freeformz, line 277: Yes it is, godep does.

bin/detect
```diff
@@ -6,9 +6,10 @@ build=$(cd "$1/" && pwd)
 
 if test -f "$build/Godeps/Godeps.json" ||
    (test -d "$build/src" && test -n "$(find "$build/src" -type f -name '*.go' | sed 1q)") || # gb info will detect any dir with a src/ dir in it as good :-()
+   (test -d "$build/vendor") || # detect vendor folder. can't check value of $GO15VENDOREXPERIMENT (go1.7 which will drop the need for $GO15VENDOREXPERIMENT)
```
freeformz, line 9: ruby apps can also have this folder.

freeformz:
I commented, just to show some of the superficial problems for posterity sake.

PR #118, opened 2016-04-11
export PATH so it's more accessbile during runtime

sandfox:
This mini change makes the`PATH` env var accessible to any sub-processes during runtime in the dyno.

### DIFF
--- a/bin/compile
+++ b/bin/compile
@@ -341,5 +341,5 @@ esac
 rm -rf $build/.heroku
 
 mkdir -p $build/.profile.d
-echo 'PATH=$PATH:$HOME/bin' > $build/.profile.d/go.sh
+echo 'export PATH=$PATH:$HOME/bin' > $build/.profile.d/go.sh
 cp $buildpack/vendor/concurrency.sh $build/.profile.d/


kr:
The environment variable PATH is already exported to begin with. This change doesn't accomplish anything.

freeformz:
@sandfox Was there a problem you were trying to solve? The unix shell sources these profile.d scripts at startup so the PATH shouldn't need to be exported.

sandfox:
Oh dear, I'm very sorry, I've been debugging the wrong buildpack..... (appears the app I'm looking at is using a very modified fork for $reasons) so I need to check if the problems we're encountering are confined to our mod/fork. 

Going to close unless/until I find out the problem is definitely related to the official buildpack

@kr just to clarify - do you mean that it's `source`d  in the way freeformz mentions or that there is something thing happening somewhere that exports it?

Some background for whats it's worth :
We're also using a bash script inside the app to run somethings which seems like it wasn't seeing `$PATH` that `.profile.d/go.sh` was setting so we're having to run commands with path to the `./bin/our-binary` rather then just `our-binary`. We noticed that some other buildpacks were `export`ing the vars inside their equivalent `.profile.d/*` and concluded that our subshell wasn't see the modified `$PATH`

bash script (very tl;dr version)

```
for file in .profile.d/*; do
    source $file
done
command="$@"
runner="bash -c"
exec $runner "$command"
```

kr:
> @kr just to clarify - do you mean that it's sourced in the way freeformz mentions or that there is something thing happening somewhere that exports it?

### Startup Process

When a dyno starts up, the heroku process manager creates its container with a basic environment (setting `$HOME` and `$PATH` etc) and executes a `bash` process with the flag `-c` and the command line from the procfile to run. As part of its startup process, bash executes ("sources") one or more shell scripts such as `/etc/profile` and `$HOME/.profile` (I don't recall the exact list). Heroku has arranged that one of these files contains a loop like the one you gave above, `for file in /app/.profile.d/*; do source $file; done` or something similar. Then, once all that startup stuff is finished, bash executes the command line it got in the `-c` flag.

### Exported Variables

It's helpful for us to distinguish between _shell variables_ and _environment variables_. Environment vars are provided by the operating system, and they're inherited by child processes when the parent calls `fork`. Shell variables exist only inside a shell script process – they are a local feature provided by bash.

Bash sort of tries to hide this distinction by exposing both shell variables and environment variables in the same namespace. So `$FOO` could be either a shell variable local to the process or an environment variable, and you can't tell which just from looking at it. But it does provide some knobs to control this. When you write `export FOO` that sort of "promotes" the variable FOO from a shell variable to be an environment variable (if it isn't already). From that point on, any changes to the value of FOO will also be copied to the environment, so for example:

```
$ FOO=x
$ env|grep FOO
$ export FOO
$ env|grep FOO
FOO=x
$ FOO=y
$ env|grep FOO
FOO=y
```

And writing `export FOO=bar` is just shorthand for `export FOO; FOO=bar`.

The last piece of the puzzle is that any variables that are already in the environment when bash starts are "exported" from the moment bash starts executing, even if you don't explicitly export them. And `PATH` is one such variable (as mentioned above). It's basically like the very first thing bash does is run `export PATH HOME FOO BAR …` to cover all the variables inherited in the environment, before executing any of the actual scripts or commands it was provided.

### Path

Putting this all together means that, in just about any bash script (inside a heroku dyno but also just about anywhere else), writing `export PATH=xyz` is no different than writing just `PATH=xyz`.

freeformz:
@kr Thanks for the detailed response.

PR #120, opened 2016-04-21
Massage the pkg spec found in Godep.json for vendor

freeformz:
Godep can save other packages other than what's local.

This is usually used to vendor one or more additional main packages like
github.com/mattes/migrate. With the old school godep workspace this just
worked because the workspace was added to the GOPATH by `godep go
install`. But with the vendor experiment we need to prefix those paths
with the `<the package>/vendor` for them to be `go install`able.

So this patch checks to see if the directory for the packages that
aren't `.*` packages exists in `<repo>/vendor` and if they do, alters
the package name to `<repo package name>/vendor/<package>`.

### DIFF
--- a/bin/compile
+++ b/bin/compile
@@ -345,8 +345,9 @@ case $TOOL in
 
         unset GIT_DIR # unset git dir or it will mess with goinstall
         cd $p
-        step "Running: go install -v ${FLAGS[@]} $pkgs"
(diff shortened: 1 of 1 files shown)


gylaz:
This has solved the issue for us, which @bernerdschaefer was helping me out with.

owenthereal:
:+1:

PR #122, opened 2016-04-27
No longer require a `Procfile` for Godeps and Govendor

RochesterinNYC:
- This makes the go buildpack run the generated main package binary name to start the web worker if no  procfile is present and no start command is specified

### DIFF
--- a/bin/compile
+++ b/bin/compile
@@ -508,6 +508,11 @@ case ${TOOL} in
     ;;
 esac
 
+if ! test -e $build/Procfile && test -n "${name}"
+then
+  echo -e "web: $(basename $name)" >> $build/Procfile
+fi
+
 rm -rf $build/.heroku
 
 mkdir -p $build/.profile.d
--- a/test/run
+++ b/test/run
@@ -388,6 +388,14 @@ testCompileGodepBasic() {
   assertFileDoesNotExist ".profile.d/concurrency.sh"
 }
 
+testCompileGodepCreateProcfile() {
+  compile "godep-basic"
+  assertCaptured "Installing go"
+  assertCapturedSuccess
+  assertCompiledBinaryExists
+  assertFile "web: fixture" "Procfile"
+}
+
 testDetectGovendorBasic() {
   detect "govendor-basic"
   assertCaptured "Go"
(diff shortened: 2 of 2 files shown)


bin/compile
```diff
@@ -384,6 +384,11 @@ case $TOOL in
     ;;
 esac
 
+if [ ! -z ${name+x} ];
```
freeformz: When will this ever be false?

bin/compile
```diff
@@ -384,6 +384,11 @@ case $TOOL in
...
+then
+  echo -e "---\ndefault_process_types:\n  web: $(basename $name)" >> /tmp/buildpack-release-step.yml
```
freeformz: This assumes $(basename $name) is the executable. This only works when $name is also the main package. I find this is rarely the case in anything except smaller apps.

freeformz:
Thanks for your interest!

Please see my comments.

Perhaps it would be better to look into bin and if there is only a single executable, use that instead, assuming there is no Procfile.

freeformz:
Also from our docs:

`Of course, rather than using default_process_types, it’s simpler to just write a default Procfile if one isn’t provided.`

I'd rather do that TBH.

bin/compile
RochesterinNYC: In the case of the gb package manager being used, it seems that `name` would not be set.

bin/compile
RochesterinNYC: The assumption that I was making is that a user would not be pushing an app without a `Procfile` unless that app is a simple/smaller app that would result in one binary (from the main package) being generated, with the intention that this generated binary would be run for the app. If the app was more complex and would result in multiple binaries being generated, then the user would include a `Procfile` to start different worker types off of the different binaries. Are there inaccuracies with this assumption?

RochesterinNYC:
Okay, I agree with the generating a default `Procfile` rather than changing `bin/release` point. I'll update the PR to reflect that.

bin/compile
```diff
@@ -384,6 +384,11 @@ case $TOOL in
     ;;
 esac
 
+if ! test -e $build/Procfile && ! test -z ${name+x}
```
freeformz: `test -n "${name}"` instead of `! test -z ${name+x}` ?

RochesterinNYC:
Updated this PR, addressed the comments, and added unit tests.

freeformz:
I am the primary maintainer and I was out on vacation. I'll look at this more soon.

freeformz:
:+1:

freeformz:
Thanks for the contribution!

PR #123, opened 2016-04-29
Tests

freeformz:


bin/compile
```diff
@@ -145,7 +150,7 @@ if [ ! -z "$env_dir" ]
 then
     mkdir -p "$env_dir"
     env_dir=$(cd "$env_dir/" && pwd)
-    ln -sfn $build /app/code
+    #
```
owenthereal: Extra?

etc/docker-test.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
```
owenthereal: `etc` usually contains configuration files in a *nix system. I understand this is to avoid messing with `bin` which is reserved for the buildpack. How about `script/docker-test`?

owenthereal:
:+1:

Are we running the tests on Travis or any CI?

etc/docker-test.sh
owenthereal: And I don't seem to find a usage of this file?

bin/compile
```diff
@@ -177,19 +181,18 @@ then
...
     then
-        warn "The 'rootPath' field not specified in 'vendor/vendor.json'."
-        warn "'rootPath' must be set to the root package name used by your repository."
-        warn "Recent versions of govendor add this field automatically, please upgrade"
-        warn "and re-run 'govendor init'."
-        warn ""
-        warn "For more details see: https://devcenter.heroku.com/articles/go-apps-with-govendor#build-configuration"
+        err 'The "rootPath" field not specified in "vendor/vendor.json".'
+        err "'rootPath' must be set to the root package name used by your repository."
+        err "Recent versions of govendor add this field a1utomatically, please upgrade"
```
tt: Should probably be "automatically" again...

bin/compile
```diff
@@ -177,19 +181,18 @@ then
...
     if [ "$name" = "null" -o -z "$name" ]
     then
-        warn "The 'rootPath' field not specified in 'vendor/vendor.json'."
-        warn "'rootPath' must be set to the root package name used by your repository."
-        warn "Recent versions of govendor add this field automatically, please upgrade"
-        warn "and re-run 'govendor init'."
-        warn ""
-        warn "For more details see: https://devcenter.heroku.com/articles/go-apps-with-govendor#build-configuration"
+        err 'The "rootPath" field not specified in "vendor/vendor.json".'
+        err "'rootPath' must be set to the root package name used by your repository."
```
tt, line 191: Should it be `"rootPath"` for consistent use of quotes?

tt:
Looks good to me (except minor comments) although I would probably write warnings to standard error.

bin/compile
```diff
@@ -145,7 +150,6 @@ if [ ! -z "$env_dir" ]
 then
     mkdir -p "$env_dir"
     env_dir=$(cd "$env_dir/" && pwd)
-    ln -sfn $build /app/code
```
msakrejda, line 148: I think this dates back to f9f79478fa74d74fb92bfdceea11080bccbea872 and 43f7b5f285ffff6ebc71f435051ea49420c1d855.

msakrejda, line 148: If we drop this, we should make sure that https://github.com/heroku/heroku-buildpack-go#using-with-cgo is still accurate or adjust it.

freeformz: Builds may not actually happen in /app

freeformz, line 148: something else to test. ;-)

freeformz:
I believe all the comments are addressed. So if there aren't any more I'll merge this in the afternoon.

PR #124, opened 2016-05-11
Sub actual build dir for literal '${build_dir}'

freeformz:
With the removal of the /app/code symlink, this allows a user to specify
paths that are relative to the build (`./` appears to not work).

I am not sure this is the _best_ way to support this, but seems better than relying on "/app"


freeformz:
I need to improve the CGO test before merging anyway.

amerine:
Seems fine to me.

freeformz:
Also, tests added.

README.md
```diff
@@ -111,11 +111,13 @@ $ make test
...
-This buildpack supports building with C dependencies via [cgo][cgo]. You can set
-config vars to specify CGO flags to specify paths for vendored dependencies. For
-example, if you added C headers to and `includes` directory, add the config var
-`CGO_CFLAGS` with the value `-I/app/code/includes` and include the relevant
-header files in `includes/` in your app.
+The buildpack supports building with C dependencies via [cgo][cgo]. You can set
+config vars to specify CGO flags to specify paths for vendored dependencies. The
+literal text of `${build_dir}` will be replaced with the directory the build is
+happening in. For example, if you added C headers to an `includes/` directory,
+add the following config to your app: `heroku config:set CGO_CFLAGS'-I${
```
msakrejda: There should be a space here after `CGO_FLAGS`, no?

freeformz: should be an `=`. Fixed. Thanks

msakrejda: Oh right you are. For some reason thought it was two separate args.

PR #125, opened 2016-05-13
Only set GOMAXPROCS defaults for old go versions

freeformz:
If we're on versions of go < 1.5 then copy in concurrency.sh, otherwise
let GO / the user determine what GOMAXPROCS should be.

This is the path of least surprise.

/cc @lstoll @fabiokung


lstoll:
This makes me super happy, ❤️

fabiokung:
:+1: :clap:

PR #130, opened 2016-05-29
Allow common go patterns

tunezaq:
Feels silly to insist that a project is go only when its using a
dependency tool. This PR would allow projects using common go file
structures to be recognised as go.

```
main.go
cmd/main.go
```

### DIFF
--- a/bin/detect
+++ b/bin/detect
@@ -6,7 +6,9 @@ build=$(cd "$1/" && pwd)
 
 if test -f "$build/Godeps/Godeps.json" || # godeps
    test -f "$build/vendor/vendor.json" || # govendor
-   (test -d "$build/src" && test -n "$(find "$build/src" -mindepth 2 -type f -name '*.go' | sed 1q)") # gb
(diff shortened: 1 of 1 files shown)


bin/detect
```diff
@@ -6,9 +6,11 @@ build=$(cd "$1/" && pwd)
 
 if test -f "$build/Godeps/Godeps.json" || # godeps
    test -f "$build/vendor/vendor.json" || # govendor
-   (test -d "$build/src" && test -n "$(find "$build/src" -mindepth 2 -type f -name '*.go' | sed 1q)") # gb
+   (test -d "$build/src" && test -n "$(find "$build/src" -mindepth 2 -type f -name '*.go' | sed 1q)") || # gb
+   test -f "$build/main.go" ||
+   test -f "$build/cmd/main.go"
```
freeformz, line 11: This would make the executable named "cmd". It's likely that we'de want to check for `$build/cmd/**/main.go` instead?

freeformz:
Missing tests.

freeformz:
There is a similar switch in compile as well, so similar bits would need to be added there as well.

Issue #143, opened 2016-08-26
glide: can't find executable


potatogopher:
I am using glide for the heroku buildpack and go1.6.3.

I get the following warnings and errors when installing. The inflect package is looking for the `hg` executable. Is the build pack not installing required executables for a package to work? Is there a work around to make this work?

```
remote: -----> Go app detected
remote: -----> Installing go1.6.3... done
remote: -----> Installing glide v0.10.2... done
remote: -----> Fetching any unsaved dependencies (glide install)
remote: [INFO] Downloading dependencies. Please wait...
remote: [INFO] Fetching updates for bitbucket.org/pkg/inflect.
remote: [INFO] Fetching updates for github.com/armon/go-metrics.
remote: [INFO] Fetching updates for github.com/denisenkom/go-mssqldb.
remote: [INFO] Fetching updates for github.com/dgrijalva/jwt-go.
remote: [INFO] Fetching updates for github.com/dimfeld/httppath.
remote: [INFO] Fetching updates for github.com/dimfeld/httptreemux.
remote: [INFO] Fetching updates for github.com/goadesign/goa.
remote: [INFO] Fetching updates for github.com/goadesign/gorma.
remote: [INFO] Fetching updates for github.com/jinzhu/gorm.
remote: [INFO] Fetching updates for github.com/jinzhu/inflection.
remote: [INFO] Fetching updates for github.com/joho/godotenv.
remote: [INFO] Fetching updates for github.com/kelseyhightower/envconfig.
remote: [INFO] Fetching updates for github.com/kr/text.
remote: [INFO] Fetching updates for github.com/lib/pq.
remote: [INFO] Fetching updates for github.com/kr/pretty.
remote: [INFO] Fetching updates for github.com/manveru/faker.
remote: [INFO] Fetching updates for github.com/satori/go.uuid.
remote: [INFO] Fetching updates for github.com/zach-klippenstein/goregen.
remote: [INFO] Fetching updates for github.com/serenize/snaker.
remote: [INFO] Fetching updates for golang.org/x/crypto.
remote: [WARN] Unable to checkout bitbucket.org/pkg/inflect
remote: [ERROR] Update failed for bitbucket.org/pkg/inflect: : exec: "hg": executable file not found in $PATH
remote: [INFO] Fetching updates for golang.org/x/net.
remote: [INFO] Fetching updates for golang.org/x/tools.
remote: [INFO] Fetching updates for gopkg.in/yaml.v2.
[...]

PR #144, opened 2016-08-26
Add hg

freeformz:


freeformz:
@potatogopher Can you try out this branch?

``` console
$ heroku buildpacks:set https://github.com/heroku/heroku-buildpack-go.git#hg
$ git push heroku master
```

freeformz:
For now I've decided to just install hg. Sometime in the future, if this becomes an issue with BZR and SVN as well I will write some code to determine VCS tools needed for the supported vendoring tools and install the needed bits.

lstoll:
:+1:

potatogopher:
@freeformz Everything works great 👍

freeformz:
I'm going to go ahead with merging and releasing this for now (so you can clear the custom BP), but I think I'm going to change things at some point to use a cached version either in the BP or dl a compiled blob from an s3 bucket to reduce external dependencies and build times.

RochesterinNYC:
@freeformz What's the context behind this PR? Why was the hg (and planned BZR) dependency added to the buildpack?

freeformz:
@RochesterinNYC If you include a dep that uses hg, `glide sync` will fail.

RochesterinNYC:
Thanks.

For public reference, issue with the context behind this PR is here: #143

freeformz:
@RochesterinNYC Ahh thanks. Sorry I should have linked it with a "Fixes #143"

PR #145, opened 2016-08-26
Merge heroku's base build pack into our custom version

egravert:


freeformz:
BTW: You can have the mainline buildpack install vendored main packages (like goose).

egravert:
@freeformz Thanks for the heads up! I am exited to be able to get back to using the standard buildpack instead of our one-off.

PR #150, opened 2016-09-19
govendor: don't sync dependencies if `.heroku.sync` == false

### LINKED ISSUE #149: govendor: can't easily manually update vendors
Hello there,

Locally to my `/vendor/` dir I am trying to do little changes to a library that _might_ have issues.

But the buildpack will always do 

```
-----> Fetching any unsaved dependencies (govendor sync)
```

which will always replace those little changes I made.

This can be avoided by committing those changes somewhere, but it'd be a pain.

I also believe code freedom inside`/vendor` is really important.

Is it possible to avoid this from happening ?
May be by adding  a `heroku.sync` bool in **vendors.json** ?

azr:
this fixes #149

### DIFF
--- a/bin/compile
+++ b/bin/compile
@@ -515,7 +515,7 @@ case ${TOOL} in
         unset GIT_DIR # unset git dir or it will mess with goinstall
         cd $p
 
-        if test `<${vendorJSON} jq -r '.heroku.sync'` != false
+        if test "$(<${vendorJSON} jq -r '.heroku.sync')" != "false"
         then
             step "Fetching any unsaved dependencies (govendor sync)"
             govendor sync
--- a/bin/compile
+++ b/bin/compile
@@ -515,8 +515,11 @@ case ${TOOL} in
         unset GIT_DIR # unset git dir or it will mess with goinstall
         cd $p
 
-        step "Fetching any unsaved dependencies (govendor sync)"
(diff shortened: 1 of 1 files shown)


azr:
probably, other dep managers need this ?

bin/compile
```diff
@@ -515,8 +515,11 @@ case ${TOOL} in
         unset GIT_DIR # unset git dir or it will mess with goinstall
         cd $p
 
-        step "Fetching any unsaved dependencies (govendor sync)"
-        govendor sync
+        if [test `<${vendorJSON} jq -r '.heroku.sync'` != false]
```
freeformz: Drop the `[` and `]`

freeformz:
Glide would need a config var. GB/godep doesn't have this type of step.

bin/compile
azr: done ! :)

azr:
okay

azr:
Hey, hem, I dont really understand if I broke the tests ?

freeformz (approved):

freeformz:
:+1:

freeformz:
Thanks for the contribution. I'll likely release this tomorrow.

azr:
Nice !! :)

PR #151, opened 2016-09-19
Optionally, install tools into image

freeformz:
If GO_INSTALL_TOOLS_IN_IMAGE is set to the string "true", then the go
toolchain and the dependency management tool is copied into the image.

The cache is still primed so we don't have to download the same go
version/dependency management tool for every build. The tools are copied
out of the cache.

This is part 1 of a few changes for upcoming feature work.


benburkert:
:+1: LGTM

amerine:
LGTM, I needed this a few days ago, heh. 👌

bin/compile
```diff
@@ -585,12 +585,25 @@ then
   echo -e "web: $(basename $name)" >> $build/Procfile
 fi
 
+cd $build
 rm -rf $build/.heroku
 
 mkdir -p $build/.profile.d
 echo 'PATH=$PATH:$HOME/bin' > $build/.profile.d/go.sh
 
+if test "${GO_INSTALL_TOOLS_IN_IMAGE}" = "true"
```
jkakar, line 594: Minor nitpick: consider using the presence of the env var as a trigger for the behaviour (ie, don't consider the value at all) to avoid confusion where someone uses `GO_INSTALL_TOOLS_IN_IMAGE=1` and then has a confusing follow on experience.

jkakar (approved):
:+1: This looks great!

bin/compile
freeformz: I tend to like explicit values for things like this as it's the most narrow contract & reduces the chances of accidental activation.

PR #153, opened 2016-09-22
$HOME/.heroku/go instead of $HOME/.go

freeformz:
This is more in line with other buildpacks (nodejs,php & python, but not jdk/ruby)


benburkert (approved):
LGTM :+1:

PR #162, opened 2016-10-21
Add GITHUB_ACCESS_TOKEN env var support for GitHub private repos

AndrasEszes:
This PR adds support for a new environment variable `GITHUB_ACCESS_TOKEN` to enable using personal access tokens for downloading dependencies from private repositories.

This allows `go get` to pull in private dependencies into Heroku applications at build time.

I use it in my personal projects but think it might be useful for everyone.


freeformz:
This could be used instead and I think it's possibly a better overall solution (haven't spent too much time thinking about this yet though): https://github.com/timshadel/heroku-buildpack-github-netrc

AndrasEszes:
Oh! :) I didn't know the netrc buildpack before now, but I seems to me better than my solution. Thank you for your reply!

freeformz:
@AndrasEszes I should probably mention it in the docs and/or the devcenter guides. I'll take a look at doing so. Thanks!

PR #165, opened 2017-01-05
Use S3 Bucket and verify files

freeformz:
This changes makes it so the buildpack downloads software only from
Heroku's S3 bucket. The SHAs of uploaded files are recorded in
files.json and those SHAs are verified on download.

Additional data about expansion, versions and what not is stored in
data.json.

Both of these changes should make it so that updating Go versions
doesn't require messing that much with the BASH scripts and should
mostly require modifications to the json files.

A new bin `sync-files.sh` is used to sync the contents of the S3 bucket
to local disk, verify the SHAs and upload new files. Deletion of files
from the bucket requires manual work however. `sync-files.sh` downloads
files to `./file-cache`, which is ignored by git.

This includes a lot of cleanups for consistency as well.

/cc @danp @lstoll @bigkevmcd @jkakar

### DIFF
--- a/.bkignore
+++ b/.bkignore
@@ -1,0 +1,1 @@
+file-cache/
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,1 @@
+file-cache/**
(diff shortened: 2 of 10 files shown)


amerine:
LGTM

PR #180, opened 2017-05-11
Install github.com/mattes/migrate/cli if specified

freeformz:
Only for govendor ATM. Probably a better way to do this, but this works for now.


freeformz:
I may actually just start pulling releases from our S3 bucket instead since there are actual releases of mattes/migrate's cli https://github.com/mattes/migrate/releases

freeformz:
Current PR pulls from our S3 bucket's idea of what version of mattes/migrate is released based on the presence of `github.com/heroku/mattes/migrate` in the `heroku.additionalTools` key of `vendor/vendor.json`.

bernerdschaefer:
Tested with the default version and it works great!

bernerdschaefer:
👍

PR #187, opened 2017-05-25
Namespace cache to prevent deleting other buildpacks' cache data

jdlehman:
Right now this buildpack will [blow away the entire cache when updating the Go version](https://github.com/jdlehman/heroku-buildpack-go/blob/7801e701b07b7bc5a599e9b7a134fce02cf34d74/bin/compile#L216). It probably isn't ideal to blow away other buildpacks' caches. 

This PR seeks to remedy this by namespacing the cache directory that this buildpack uses by putting it in `$cache/go-cache`. The best part about this change is that it is small and shouldn't break anything.

### DIFF
--- a/bin/compile
+++ b/bin/compile
@@ -4,8 +4,9 @@
 set -eo pipefail
 
 mkdir -p "$1" "$2"
+mkdir -p "$2/go-cache"
 build=$(cd "$1/" && pwd)
-cache=$(cd "$2/" && pwd)
+cache=$(cd "$2/go-cache" && pwd)
 env_dir="${3}"
 
 buildpack=$(cd "$(dirname $0)/.." && pwd)


bin/compile
```diff
@@ -4,8 +4,9 @@
 set -eo pipefail
 
 mkdir -p "$1" "$2"
+mkdir -p "$2/go-cache"
 build=$(cd "$1/" && pwd)
-cache=$(cd "$2/" && pwd)
```
freeformz, line 8: A check to ensure that `$2/go-cache` exists need to happen before the subshell, otherwise cache is set to ""

```console
$ echo "'$(cd nonexistant && pwd)'"
cd: no such file or directory: nonexistant
''
```

freeformz:
There is also an argument to be made that we don't know what other buildpacks the user may have built that uses the version of go that was in the cache and we now need to nuke those buildpack's caches as well so they an rebuild properly. I'm still of the opinion that if a major piece changes (node version, go version, ruby version, etc) the sanest thing to do is consider all bets off and nuke the cache.

jdlehman:
Thanks for reviewing this. I'd make the argument that it is the individual buildpack's job to handle its own cache. If a buildpack is relying on another buildpack as a dependency, let's say Go, and the Go version changes, I'd assume that it would be that buildpack's responsibility to recognize that it's cached data was built with a now changed dependency and to rebuild correctly.

That said, it is definitely easier to assume that other buildpacks might not be managing their caches effectively in this manner and just nuke the entire cache. If I have a machine that runs all of my builds, and if some of my apps use different versions of Go, it feels wrong to blow away the entire cache every time one app relies on a different version of Go. To the best of my knowledge, I don't believe I know of any other buildpack that takes this approach. I believe that both the ruby and node buildpacks handle installing multiple versions without blowing away the entire cache (that said, I am no expert and might be incorrect).

Ultimately I think it is up to you. I can make the request changes if we want to avoid nuking the entire cache in this instance. However, if the Heroku vision was to isolate the caches in the way I describe, it would be easy enough to create a new cache directory per buildpack and pass that directory along to `compile`, so maybe this PR goes against the spirit and vision of these buildpacks.

freeformz:
ATM (I don't know if this will always be the case) the cache directory passed to buildpacks is shared and is per app. I understand your PoV and don't disagree, I'm just erring more on the side of caution (where caution == blow everything away, caches aren't meant to be permanent). But I can see how, for instance, an upgrade to the go version, blowing away the node cache (for an app that uses both buildpacks) just makes that build take forever, is unexpected and makes for a bad user experience.

I am also wondering if setting the cache this way is the best option as well. Maybe it's better to root everything under the $goVersion and we can then look for an older $goVersion and blow those away. This way everything compiled with that version of go ends up being rooted in the same place and get's blown away only after a successful compile with the new version of go? This would move the cleanup of the cache to end of the buildpack, making a downgrade after a failure not have to re-cache everything. I'm also not sure that's important either. ;-) Thoughts?

jdlehman:
Ah, I didn't know the cache was per app. That makes much more sense. I think rooting everything under the $goVersion isn't a bad idea, but if the cache is meant to be per app, then maybe this isn't as big of an issue as I initially thought.

I am fine closing this unless you think rooting under $goVersion is worth the effort.

freeformz:
Let's close it then. If it's a problem others will complain and it can be revisited.

PR #210, opened 2017-09-26
Dep Support

freeformz:
Should be complete including testpack support.


bin/compile
```diff
@@ -357,6 +383,29 @@ export GOPATH
...
+        if [ -z "${pkgs}" ]; then
+            pkgs="default"
+        fi
+        warnPackageSpecOverride
+        handleDefaultPkgSpec
+
+        unset GIT_DIR # unset git dir or it will mess with goinstall
+        cd "${src}"
+
+        if [ "$(< ${depTOML} tq '$.metadata.heroku["ensure"]')" != "false" ]; then
```
owenthereal, line 399: What do you think of only running `dep ensure` if the manifest declares it true?

owenthereal: I meant by default we encourage people to push dependencies to the repo and what's in the repo is used hence `dep ensure` is not usually needed.

freeformz: Since it's the opposite of how the other managers work (govendor / glide) I didn't really consider it.

For correctness sake I'd err on running dep ensure, AFAIK it doesn't do anything network wise if the inputs-digest of the code matches whats in the repo, so for anyone who checks in the contents of vendor/ it's quick and just some disk I/O. For those that don't do check in vendor/, or forget to add the updated deps in vendor/ then running `dep ensure` saves them from themselves (or at least tries to).

amerine (approved):

PR #217, opened 2017-10-13
Allow `go1.X.0` versions that expand to `go1.X`.

freeformz:
This allows a user to pin to the first released version of a series
and would not be automatically upgraded to later versions.

Example:

A go version of `go1.9.0` would pin the app to `go1.9`.


jkakar (approved):
Looks good!

owenthereal (approved):

PR #218, opened 2017-10-16
publish: Publish from temp directory

freeformz:
This is easier and means we can put exactly what we want to publish
into the tmp directory instead of moving stuff we don't want to publish
out of the way.

This is also a good time to cleanup the unused files.


amerine (approved):
Nice

PR #224, opened 2018-01-08
README: fix app create command

kevinburke:


kevinburke:
Ugh, this isn't right either. I think something's broken in my local environment.

PR #241, opened 2018-03-10
Update compile

bt:


freeformz:
FWIW: There is no guarantee that the buildpack is working on a checkout of the git repo. AFAIK it's most often not, if at all.

bt:
Hey @freeformz sorry I actually pointed this PR the wrong direction; it was meant to be at my own master branch.

I've forked the buildpack to enable me to pass git versioning via envvar then into `ldflags`. Didn't turn out as good as I wanted, but I managed to integrate it with my CI by passing the variables via `config:set` so all good now :)

PR #243, opened 2018-03-30
README: clarify stdlib only projects

Xe:
Using the directions for other vendoring systems also works for projects that only use packages found in the standard library.

### DIFF
--- a/README.md
+++ b/README.md
@@ -190,7 +190,7 @@ following contents, adjusted as needed for your project's root path.
 
 ```json
 {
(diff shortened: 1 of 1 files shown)


freeformz:
@Xe Can you change the link in the comment section of that json to `https://devcenter.heroku.com/articles/go-support`. Thanks!

freeformz (changes requested):
Please change the link in the comment section of the json to point to: https://devcenter.heroku.com/articles/go-support

freeformz (approved):

PR #245, opened 2018-04-13
bin/compile: set GOCACHE inside cache directory for build caching

danp:
Set `GOCACHE` to Go 1.10+


bin/compile
```diff
@@ -294,6 +295,12 @@ ensureGo() {
     fi
 
     export GOROOT="${goPath}"
+
+    # Export GOCACHE if Go >= 1.10
+    if [ "$ver_major" -ge 1 ] && [ "$ver_minor" -ge 10 ]; then
```
danp: Should this be something like `( ver_major == 1 && ver_minor >= 10 ) || ver_major > 1`?

owenthereal: Ah, good catch! Fixed with `--amend`

bin/compile
```diff
@@ -294,6 +295,12 @@ ensureGo() {
     fi
 
     export GOROOT="${goPath}"
+
+    # Export GOCACHE if Go >= 1.10
+    if [ "$ver_major" -eq 1 ] && [ "$ver_minor" -ge 10 ] || [ "$ver_major" -gt 1 ]; then
```
danp: Maybe it would help to have parens around the first two conditions to make the grouping clearer? At least I'm not very familiar with how ordering works here.

freeformz:
Thoughts on using `semver compare 1.10 $(go version | sed -e 's/.*go\([0-9]\.[0-9]\{1,2\}\(\.[0-9]\{1,2\}\)*\).*/\1/')` from https://github.com/fsaintjacques/semver-tool and listing [one of the releases](https://github.com/fsaintjacques/semver-tool/releases) in files.json instead of having to vendor semver bash?

bin/compile
```diff
@@ -362,6 +369,12 @@ trap clearGitCredHelper INT TERM EXIT
 
 determineTool
 
+ver_major=0
```
freeformz: Please use goVerMajor, goVerMinor, goVerPatch, goVerSpecial as it's more consistent with the rest of the variables in the buildpack.

bin/compile
```diff
@@ -362,6 +369,12 @@ trap clearGitCredHelper INT TERM EXIT
...
+ver_minor=0
+ver_patch=0
+ver_special=""
+semverParseInto "$ver" ver_major ver_minor ver_patch ver_special
```
freeformz: This needs to be done after the expandVer below otherwise you get the requested version not the final one. That probably doesn't make a difference now, but it could in the future.

test/utils
```diff
@@ -121,6 +121,16 @@ assertFileExists() {
   assertTrue "File ${name} does not exist" "[ -f ${tgt} ]"
 }
 
+assertDirExists() {
+  local path=${1}
+  assertTrue "Dir ${path} does not exist" "[ -d ${path} ]"
```
freeformz: s/does not//
s/exist/exists/

test/run
```diff
@@ -969,8 +969,46 @@ testGodepPackageSpec() {
...
+
+  compile
+  assertCaptured "Installing go1.10.1"
+  assertCaptured "Fetching dep"
+  assertCaptured "Fetching any unsaved dependencies (dep ensure)"
+  assertCaptured "Running: go install -v -tags heroku ."
+  assertCaptured "github.com/heroku/fixture"
+  assertCapturedSuccess
+  assertCompiledBinaryExists
+  assertDirExists $cache_dir
```
freeformz: Do we want to assert that something specific made it's way into the dir as well?

bin/copy
```diff
@@ -21,6 +21,7 @@ function ctt {
 
 mkdir -p ${TMP}/bin
 mkdir -p ${TMP}/lib/src/github.com/apg/patter/
+mkdir -p ${TMP}/lib/semver_bash/
```
freeformz: Maybe we should put this in `vendor/` instead? i.e. it's a lib we're vendoring, not one we've written.

bin/compile
freeformz: Also, I think we can use something like `if semverGT 1.10 ${ver}; then .... fi` instead of all the variables?

freeformz (changes requested):

test/utils
owenthereal: You found issue in the right direction but the one had issue is line 131 down there. I fixed it in https://github.com/heroku/heroku-buildpack-go/pull/245/commits/570e7964393f0464002ea59b059863c4c4b6d8e1

owenthereal:
@danp @freeformz PTAL again. I should have addressed all concerns.

owenthereal:
Doh, getting the error when comparing a string to an integer in build: `ASSERT:Expected STD_ERR to be empty; was </buildpack/bin/compile: line 305: [: devel-15f7a66: integer expression expected> expected:<> but was:</buildpack/bin/compile: line 305: [: devel-15f7a66: integer expression expected>`. Not sure what's the cleanest way to fix it.

danp:
Maybe different approach: what if we base setting GOCACHE on whether it shows up in `go env`? That should work with devel versions too.

owenthereal:
@danp:

> Maybe different approach: what if we base setting GOCACHE on whether it shows up in go env? That should work with devel versions too.

❤️ the solution and it's sooo much cleaner now!

bin/compile
```diff
@@ -295,6 +295,12 @@ ensureGo() {
 
     export GOROOT="${goPath}"
     PATH="${goPath}/bin:${PATH}"
+
+    # Export GOCACHE if Go >= 1.10
+    local goCachePath=$(go env | grep GOCACHE)
```
danp: I think this could be: `if go env | egrep -q '^GOCACHE='; then`.

owenthereal: TIL. My bash skill makes me mad 😝

owenthereal: Fixed

danp:
LGTM! Maybe the troels/heroku:18-build image has gone away.

danp:
Could maybe use a changelog entry as well?

owenthereal:
> Could maybe use a changelog entry as well?

Done

owenthereal:
Not sure why went wrong. The same test `testGoCacheGoVersionGreaterThanOrEqualTo110` passed on my machine...

danp:
Would be nice if we could see the output. Maybe change things up to temporarily dump it so we can see what trouble travis is running into?

owenthereal:
It's weird that my debugging output doesn't show up in STDIN/STDERR on Travis. I was able to see them locally though...

owenthereal:
@freeformz @danp After bumping the fixture from 1.10.1 to 1.10.2 then the `testGoCacheGoVersionGreaterThanOrEqualTo110` passed on Travis. Super strange...but anyways. this is ready to be merged I think.

danp:
Hmm, wonder if it will break again once, say, 1.10.3 comes out.

@freeformz want to redo your review?

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 * Add go1.10.2 and go1.9.6
 * Default to go1.10.2
 * go1.10 expands to go1.10.2 and go1.9 expands to go1.9.6
+* GOCACHE support
```
alindeman: I think this entry should be in `Unreleased` until a release is cut.

alindeman (approved):
Very nice! I am looking forward to having this available.

CHANGELOG.md
owenthereal: Good find. Fixed.

owenthereal:
@freeformz Would you mind looking at this?

test/fixtures/dep-go-1-10/Gopkg.toml
```diff
@@ -0,0 +1,25 @@
+[metadata.heroku]
+  root-package = "github.com/heroku/fixture"
+  go-version = "go1.10.2"
```
freeformz: change this to `go1.10` and it will expand to the most recent version of go1.10.X

bin/compile
```diff
@@ -295,6 +295,11 @@ ensureGo() {
 
     export GOROOT="${goPath}"
     PATH="${goPath}/bin:${PATH}"
+
+    # Export GOCACHE if Go >= 1.10
+    if go env | egrep -q '^GOCACHE='; then
```
freeformz: please switch to `grep`.

bin/compile
```diff
@@ -295,6 +295,11 @@ ensureGo() {
...
+      export GOCACHE="${cache}/${goVersion}/build-cache"
```
freeformz: What is the default and why is that not sufficient?

freeformz (changes requested):

test/fixtures/dep-go-1-10/Gopkg.toml
owenthereal: Fixed with `--amend`

bin/compile
owenthereal: Fixed with `--amend`

bin/compile
owenthereal: By default `GOCACHE="/root/.cache/go-build"` which isn't carried forward between builds.

freeformz:
To answer my own question above the default is `GOCACHE="/app/.cache/go-build"`, which is :-(

freeformz (approved):

test/fixtures/dep-go-1-10/Gopkg.toml
owenthereal: I just realized we couldn't use `go1.10` here since the test is asserting a folder having the specific go version there: https://github.com/heroku/heroku-buildpack-go/pull/245/commits/f4a7f0a2bd7d883ecfae37fc0942450e21c34551#diff-a6e781883513014286df21a2e0e8aaa2R973. If it's `go1.10` then the test will have to be updated on every minor version update.

freeformz:
@jingweno Suggestion: set the cache to `${cache}/go-build-cache` and it will be cleared during any go version change. This makes the test(s) stable as well.

freeformz:
Thanks!

freeformz:
Released as [v89](https://github.com/heroku/heroku-buildpack-go/releases/tag/v89)

danp:
Thanks so much for sticking with this, @jingweno!

PR #251, opened 2018-06-12
Add / default to go1.10.3 and go1.9.7

freeformz:
* Add go1.10.3 and go1.9.7
* Default to go1.10.3
* go1.10 expands to go1.10.3 and go1.9 expands to go1.9.7


sowjumn:
looks good to me.

amerine (approved):

PR #257, opened 2018-08-02
Move from github.com/mattes/migrate => github.com/golang-migrate/migrate repo

### LINKED ISSUE #256: Migrate `mattes/migrate` to new repo `golang-migrate/migrate`
The `mattes/migrate` repo is archived since 8th May and superseded by [golang-migrate/migrate](https://github.com/golang-migrate/migrate) now. Support for this repo instead (or mid-term both side-by-side to not break current app deployments) would be nice to keep up with features and security patches ([v3.0.1](https://github.com/mattes/migrate/releases/tag/v3.0.1) vs [v.3.4.0](https://github.com/golang-migrate/migrate/releases/tag/v3.4.0])).

Of course, one can fork the buildpack or use install it as a vendored `go get` dependency, but would be nice to maintain that feature.

I might give the PR a try, but seeing how the repo patch is regexp in different tools (`sed`, `grep`, `jq`, `tq`) will likely make supporting both more complex that the benefit it offers.

djui:
Fixes https://github.com/heroku/heroku-buildpack-go/issues/256.

Also fixes the filename download URL: `migrate.linux-amd64.tar.gz` not `migrate-linux-amd64.tar.gz`.

### DIFF
--- a/bin/compile
+++ b/bin/compile
@@ -185,8 +185,8 @@ ensureGolangMigrate() {
     else
         rm -rf "$(dirname "${mmPath}")"
         mkdir -p "${mmPath}"
-        ensureFile "migrate-${mmVersion}-linux-amd64.tar.gz" "${tmp}" "tar -C ${mmPath} -zxf"
         step "Installing github.com/golang-migrate/migrate ${mmVersion}"
+        ensureFile "migrate-${mmVersion}.linux-amd64.tar.gz" "${tmp}" "tar -C ${mmPath} -zxf"
         mv "${mmPath}/migrate.linux-amd64" "${targetFile}"
         chmod a+x "${targetFile}"
     fi
--- a/bin/compile
+++ b/bin/compile
@@ -19,7 +19,7 @@ GovendorVersion="$(<${DataJSON} jq -r '.Govendor.DefaultVersion')"
 GBVersion="$(<${DataJSON} jq -r '.GB.DefaultVersion')"
 PkgErrorsVersion="$(<${DataJSON} jq -r '.PkgErrors.DefaultVersion')"
 MercurialVersion="$(<${DataJSON} jq -r '.HG.DefaultVersion')"
-MattesMigrateVersion="$(<${DataJSON} jq -r '.MattesMigrate.DefaultVersion')"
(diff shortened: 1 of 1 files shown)


danp:
Thanks for this!

Looks like some test output checking needs to be updated.

Also, what should folks who still have github.com/mattes/migrate in heroku.additionalTools expect to happen?

djui:
>Looks like some test output checking needs to be updated.

Was not aware of the test cases. Will take care of it.

>Also, what should folks who still have github.com/mattes/migrate in heroku.additionalTools expect to happen?

That's a valid question. I am not sure. So far I would assume that no-one uses that feature as it actually does not work. So we might not have to worry about that case. But likely only you know the numbers for that.

danp:
We do use it internally. Re it not working based on filename, maybe there is confusion between the filename key here (which I think `ensureFile` eventually works from) and the released filename:

https://github.com/heroku/heroku-buildpack-go/blob/8f0649cbba8264a47faccd25257f326aede0a3a3/files.json#L334-L337

Perhaps the buildpack should support either being in heroku.additionalTools, preferring github.com/golang-migrate/migrate, and printing a warning if only github.com/mattes/migrate is used?

danp:
Should add too there are tests that specifically verify having github.com/mattes/migrate in heroku.additionalTools works, such as:

https://github.com/heroku/heroku-buildpack-go/blob/8f0649cbba8264a47faccd25257f326aede0a3a3/test/run#L30-L46

and:

https://github.com/heroku/heroku-buildpack-go/blob/8f0649cbba8264a47faccd25257f326aede0a3a3/test/run#L159-L176

freeformz:
This won't work as it breaks existing users who could push afterwards and not get mattes migrate as they expected.

freeformz:
See #264 for how I would like to do it so as to not break existing users and allow them some time to migrate.

freeformz:
I'm going to close this since I merged #264. @djui Thanks for bringing up this issue. I'm sorry we couldn't use your PR in the end.

djui:
@freeformz makes sense and thanks for taking over this issue! 🙇

PR #259, opened 2018-08-15
bin: add add-version to make files.json adds easier

danp:
Adds `bin/add-version` to do the work of finding a release's sha256 and adding to `files.json`.

Uses `jq -S` so adds a commit which applies that to `files.json` to start.


alindeman (approved):
Nice!

owenthereal:
❤️

PR #261, opened 2018-08-18
Basic Module Support

freeformz:
The plan is to start here and add test cases/fixes for other uses as things progress.

### DIFF
--- a/bin/compile
+++ b/bin/compile
@@ -400,6 +400,21 @@ export GOPATH
 
 # GB installation
 case "${TOOL}" in
+    gomodules)
+        pkgs=${GO_INSTALL_PACKAGE_SPEC:-$(awk '{ if ($1 == "//" && $2 == "+heroku" && $3 == "install" ) { print substr($0, index($0,$4)) }}' ${goMOD})}
+        if [ -z "${pkgs}" ]; then
+            pkgs="default"
+        fi
+        warnPackageSpecOverride
+        handleDefaultPkgSpec
+        massagePkgSpecForVendor
+
+        unset GIT_DIR # unset git dir or it will mess with goinstall
+        cd ${build}
+        export GOBIN="${build}/bin"
+        step "Running: go install -v ${FLAGS[@]} ${pkgs}"
+        go install -v "${FLAGS[@]}" ${pkgs} 2>&1
+    ;;
     dep)
         eval "$(setupGOPATH ${name})"
         depTOML="${src}/Gopkg.toml"
--- a/bin/compile
+++ b/bin/compile
@@ -33,6 +33,13 @@ handleDefaultPkgSpec() {
         warn "Installing package '.' (default)"
         warn ""
         case "${TOOL}" in
+            gomodules)
+                warn "To install a different package spec add a comment in the following form to your \`go.mod\` file:"
+                warn "// +heroku install ./cmd/..."
+                warn ""
+                warn "For more details see: https://devcenter.heroku.com/articles/go-apps-with-modules#build-configuration"
+                warn ""
+            ;;
             dep)
                 warn "To install a different package spec set 'metadata.heroku.pkg-spec' in 'Gopkg.toml'"
                 warn ""
--- a/bin/compile
+++ b/bin/compile
@@ -416,6 +416,10 @@ case "${TOOL}" in
         handleDefaultPkgSpec
         massagePkgSpecForVendor
 
+        if [ -d "${build}/vendor" -a -s "${build}/go.sum" ]; then
+            FLAGS=(${FLAGS[@]} -mod=vendor)
+        fi
+
         unset GIT_DIR # unset git dir or it will mess with goinstall
         cd ${build}
         export GOBIN="${build}/bin"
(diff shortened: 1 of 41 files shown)


lib/common.sh
```diff
@@ -267,7 +268,11 @@ setGoVersionFromEnvironment() {
 }
 
 determineTool() {
-    if [ -f "${depTOML}" ]; then
+    if [ -f "${goMOD}" ]; then
+        TOOL="gomodules"
+        ver=${GOVERSION:-$(awk '{ if ($1 == "//" && $2 == "+heroku" && $3 == "goVersion" ) { print $4 } }' ${goMOD})}
```
owenthereal: Should we exit `awk` after the first match? If users mistaken have multiple `// +heroku goVersion 1.10` there will be multiple returns:

```
awk '{ if ($1 == "//" && $2 == "+heroku" && $3 == "goVersion" ) { print $4; exit; } }' ${goMOD}
```

bin/compile
```diff
@@ -400,6 +407,21 @@ export GOPATH
 
 # GB installation
 case "${TOOL}" in
+    gomodules)
+        pkgs=${GO_INSTALL_PACKAGE_SPEC:-$(awk '{ if ($1 == "//" && $2 == "+heroku" && $3 == "install" ) { print substr($0, index($0,$4)) }}' ${goMOD})}
```
owenthereal: Should we exit `awk` after the first match? If users mistaken have multiple lines of `// +heroku install ./cmd/...` there will be multiple returns:

```
awk '{ if ($1 == "//" && $2 == "+heroku" && $3 == "install" ) { print substr($0, index($0,$4)); exit; }}' ${goMOD}
```

test/run
```diff
@@ -1,6 +1,46 @@
...
+
+  assertDetected
+
+  compile
+  assertCaptured "Installing go"
+  assertCaptured "Running: go install -v -tags heroku ."
+  assertCaptured "github.com/heroku/fixture"
+  assertCapturedSuccess
+  assertCompiledBinaryExists
+}
```
owenthereal: Would it be possible to add a test for `// +heroku install xxx`?

lib/common.sh
```diff
@@ -267,7 +268,11 @@ setGoVersionFromEnvironment() {
 }
 
 determineTool() {
-    if [ -f "${depTOML}" ]; then
+    if [ -f "${goMOD}" ]; then
```
owenthereal, line 271: Do we also need to detect if `GO111MODULE=on` to enable gomodules?

test/fixtures/mod-deps-vendored/main.go
```diff
@@ -0,0 +1,11 @@
+package main
+
+import (
+	"fmt"
+
+	"github.com/gorilla/mux"
```
jingweno: Hmm...interesting. How did this test pass? Without specifying `GO111MODULE=on`, Go1.11rc2 doesn't fetch deps for me.

jingweno: NVM, you vendored all deps for this fixture.

test/run
jingweno: Also it'd be nice to have a test that Go1.11 will download modules if they are not vendored.

test/fixtures/mod-deps/main.go
```diff
@@ -0,0 +1,11 @@
+package main
+
+import (
+	"fmt"
+
+	"github.com/gorilla/mux"
```
owenthereal, line 6: Hmm...interesting. How did this test pass? Without specifying GO111MODULE=on, Go1.11rc2 doesn't fetch deps for me.

freeformz: Are you using modules inside of a GOPATH? If so that's why. AFAICT, outside of a GOPATH, modules are turned on.

lib/common.sh
freeformz: Good catch.

test/fixtures/mod-deps/main.go
owenthereal: Ah, that's true. I just verified it. I'm not sure I follow the rationale behind this though (I would expect it's either on or off).

So I assume GOPATH hasn't been set in our case that's why the test passes?

test/run
freeformz: Done.

test/fixtures/mod-deps/main.go
freeformz: Yes, when a go.mod file is detected, we don't construct a $GOPATH, we just figure some stuff out and build the code. If this hold true I can't wait for the day when I can remove all the other tools. :-)

lib/common.sh
freeformz: I'd rather not until and issue is raised or we understand better the circumstances around needing to force it on.

test/fixtures/mod-deps/main.go
owenthereal: ❤️

owenthereal (approved):

PR #263, opened 2018-08-24
Add some module safeguards / info

freeformz:
With this you should get a better error in case you are trying to use modules with a version of Go that we know doesn't support modules.

### DIFF
--- a/bin/compile
+++ b/bin/compile
@@ -25,7 +25,7 @@ TQVersion="$(<${DataJSON} jq -r '.tq.DefaultVersion')"
 
 # For specified versions of Go we need to keep concurrency.sh
 needConcurrency() {
-    <"${DataJSON}" jq -e '.Go.NeedsConcurrency | indices(["'${1}'"]) | length | if . == 0 then false else true end' &> /dev/null
+    <"${DataJSON}" jq -e '.Go.NeedsConcurrency | any(. == "'${1}'")' &> /dev/null
 }
 
 handleDefaultPkgSpec() {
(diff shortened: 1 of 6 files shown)


lstoll (approved):

bin/compile
```diff
@@ -108,10 +108,33 @@ expandVer() {
...
+        warn ""
+        warn "Go modules are an experimental feature of go1.11"
+        warn "Any issues building code that uses Go modules should be"
+        warn "reported via: https://github.com/heroku/heroku-buildpack-go/issues"
+        warn ""
+        warn "Additional documentation for using Go modules with this buildpack"
+        warn "can be found here: https://github.com/heroku/heroku-buildpack-go#go-module-specifics"
+        warn ""
+        if ! <"${DataJSON}" jq  -e '.Go.SupportsModuleExperiment | any(. == "'${ver}'")' &> /dev/null; then
+            err "You are using ${ver}, which does not support the Go modules experiment"
```
owenthereal, line 123: `which does not support` => `which does not enable`?

data.json
```diff
@@ -39,6 +39,10 @@
     "SupportsVendorExperiment": [
       "go1.5", "go1.5.1", "go1.5.2", "go1.5.3", "go1.5.4",
       "go1.6", "go1.6.1", "go1.6.2", "go1.6.3", "go1.6.4"
+    ],
+    "SupportsModuleExperiment": [
+      "go1.11beta2", "go1.11beta3", "go1.11rc1", "go1.11rc2",
+      "go1.11"
```
owenthereal, line 45: It would be nice to make it scalable for go1.11+, e.g. symver parse the version etc.

bin/compile
freeformz: IMO `support` is a more accurate word than `enable` here. If we've gotten here then we've already detected a `go.mod` file and selected go modules as the way to do the install, but either because of the release state of go1.11 or their specified go version they/the buildpack has chosen a version of go that does not support go modules. The simplest thing is for us to error out. We *could* attempt to re-run tool detection w/o gomodule support and start again but that would be a really large change and I'd prefer not to have to do that unless we can prove it's necessary.

data.json
freeformz: I have thought the same thing, but both semver and arrays like these are both solutions to the same problem. Since it will only be an experiment for several versions I have been hesitant to integrate a semver bash library into the buildpack. It may be worth considering though. If you would like to PR an implementation like that I'd love to review it to see if/how it makes this and similar structures easier to maintain in the buildpack.

bin/compile
owenthereal: :+1:

PR #267, opened 2018-08-30
Adjust version info

### LINKED ISSUE #266: error when trying to use the latest release of go with modules
The readme for module mode says "If not specified defaults to the most recent supported version of Go."

I tried to deploy an app this way, but got an error message:

```
remote: -----> Go app detected
remote:  !!    
remote:  !!    Go modules are an experimental feature of go1.11
remote:  !!    Any issues building code that uses Go modules should be
remote:  !!    reported via: https://github.com/heroku/heroku-buildpack-go/issues
remote:  !!    
remote:  !!    Additional documentation for using Go modules with this buildpac
remote:  !!    can be found here: https://github.com/heroku/heroku-buildpack-go#go-module-specifics
remote:  !!    
remote:  !!    You are using , which does not support the Go modules experiment
remote:  !!    
remote:  !!    Please add the following comment to your go.mod file to specify go1.11:
remote:  !!    // +heroku goVersion go1.11
remote:  !!    
remote:  !!    Then commit and push again.
```

I can prepare a reduced testcase if that would help.

freeformz:
Closes #266

Basically the buildpack wasn't entirely accurate when it said "the latest supported version", when in reality it uses the DefaultVersion specified in data.json when no version is specified.

Additionally the version message handling wasn't in the right place for go modules.

And when detecting go.mod, default to go1.11 for now.

This PR fixes those two things.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 * Be clearer about what version of go is chosen if none is specified. Addresses #266.
+* Handle version stuff in the right place for go modules.
 
 ## v92 (2018-08-27)
 * Add go1.11 and mark it as supported
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Go Buildpack Changelog
 
 ## Unreleased
+* Be clearer about what version of go is chosen if none is specified. Addresses #266.
 
 ## v92 (2018-08-27)
 * Add go1.11 and mark it as supported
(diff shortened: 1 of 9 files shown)


danp:
Should the go.mod/etc detection path start you with go1.11 now instead of needing to have it specified in go.mod?

freeformz:
I don't feel comfortable making that assumption. I can't really explain why though either.

freeformz:
@danp Actually nevermind. thinking about this it probably makes the most sense for now.

danp (approved):

danp:
Nice, I like it since it lets people avoid the "have to specify go1.11 now, forget to bump for go1.12+ later" issue.

PR #268, opened 2018-09-17
Fix go modules support for projects without explicit Procfile

sponomarev:
The name of the binary is determined by a module name from `go.mod` file.

Without this change, go modules support is inconsistent with the rest dependency toolchains. It doesn't support implicit Procfile creation like for instance `dep`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+* Remove the need for Procfiles in simple situations for go modules
+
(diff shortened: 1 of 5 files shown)


freeformz:
@sponomarev Thanks. Can you please add to bullet point to the changelog as well? Once that's done I can merge this.

Thanks!

sponomarev:
@freeformz Done. I hope the wording is good enough.

freeformz:
Thanks!

PR #269, opened 2018-10-03
add go1.11.1, set as default for go1.11

danp:


owenthereal (commented):
Build fail due to github.com/go-test-user/go-private-repo-test is gone. LGTM otherwise.

freeformz:
github.com/go-test-user/go-private-repo-test isn't gone. it's a private repo as that's part of the test.

PR #272, opened 2018-10-26
Fix default pkg spec warning message when using dep

andrewslotin:
`bin/compile` retrieves the list of Go packages to install from `metadata.heroku['install']` section of `Gopkg.toml`, however the default value warning message suggests that it should be set in
`metadata.heroku['pkg-spec']`. This PR changes this message to use `install` instead.

### DIFF
--- a/bin/compile
+++ b/bin/compile
@@ -42,7 +42,7 @@ handleDefaultPkgSpec() {
                 warn ""
             ;;
             dep)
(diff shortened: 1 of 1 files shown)


andrewslotin:
@freeformz, is there any chance to have this merged?

freeformz:
Thanks!

PR #273, opened 2018-11-03
Add go1.11.2 (go1.11 default) / go1.10.5 (go1.10 default)

freeformz:


files.json
```diff
@@ -80,6 +80,10 @@
     "SHA": "fa1b0e45d3b647c252f51f5e1204aba049cde4af177ef9f2181f43004f901035",
     "URL": "https://dl.google.com/go/go1.10.3.linux-amd64.tar.gz"
   },
+  "go1.10.5.linux-amd64.tar.gz": {
```
danp: Minor: looks like these aren't sorted, such as if bin/add-version was not used, so next time someone does use bin/add-version it will produce a larger diff.

danp (approved):

files.json
freeformz: Fixed

PR #275, opened 2018-11-14
Fix gitCredHelper function for go module usage

chrisroberts:
Since the =~ is a bash extension, it requires double brackets
to be evaluated properly. Using only single brackets results in
a syntax error.

Could also be updated to use:

```bash
if printf "${t}" | grep -Eq ":" ; then
```

if POSIX compatibility is important. Without this update trying to 
build with go modules and private repositories fails:

> }; gitCredHelper get: 46: [: MY_GITHUB_TOKEN: unexpected operator


freeformz:
Thanks!

PR #285, opened 2019-02-05
Run bin/pre-compile and/or bin/post-compile if they're -x

freeformz:
Fixes #78 (sort of)

78 was initially about running migrations, which shouldn't be done
during a build. For that use release phase.

But it's useful to be able to do other things when the go tools are
around without having to install them into your slug (see
GO_INSTALL_TOOLS_IN_IMAGE).


freeformz:
cc @jmorrell

bin/compile
```diff
@@ -452,8 +452,16 @@ case "${TOOL}" in
         unset GIT_DIR # unset git dir or it will mess with goinstall
         cd ${build}
         export GOBIN="${build}/bin"
+        if [[ -x bin/pre-compile ]]; then
```
jmorrell: If this file exists, but is not executable could we show a helpful message about needing to `chmod +x` it and maybe add a shebang?

I'm also unfamiliar with what this would require for Windows users.

jmorrell: Prior art: Python will explicitly `chmod +x` the file

https://github.com/heroku/heroku-buildpack-python/blob/master/bin/steps/hooks/post_compile#L3

Is there any downside to this other than it being less explicit?

README.md
```diff
@@ -80,6 +80,10 @@ greater than zero, `go install` is invoked with `-mod=vendor`, causing the build
 to skip downloading and checking of dependencies. This results in only the
 dependencies from the top level `vendor` directory being used.
 
+If the file `bin/pre-compile` or `bin/post-compile` exists in the repository and
```
jmorrell: I have been a little confused by the buildpack putting binaries into `/bin` when my local build puts it in `/` by default. Is this a go-ism that I'm unfamiliar with? cc @hunterloftis 

Does this cause a conflict in the (highly unlikely) case of the user building a binary named `pre-compile` or `post-compile`?

At the risk of bike-shedding, is there a good argument for `bin/post-compile` over `./post-compile` or `heroku/post-compile`? or is this an arbitrary decision?

bin/compile
CaseyFaist: ++ on explicit `chmod`, if we do it for them that's one less step our users have to remember in order to use this functionality and leads to better experiences

bin/compile
```diff
@@ -452,8 +452,16 @@ case "${TOOL}" in
...
+            echo "-----> Running bin/pre-compile hook"
+            sub_env bin/pre-compile
```
CaseyFaist: `sub_env` hooks in the Python buildpack have been a problem because it has unexpected consequences. Users have tried setting env vars, for example, and not known why they're not affecting build

CaseyFaist:
I read through the linked PR, it seems like we're trying to address a lot of functionality via one mechanism and deriving some guidelines on that from the python buildpack - I don't necessarily recommend following that particular pattern though, as `sub_env` has unexpected outcomes when users try to use the hooks, and at least in Python was a broad stroke that tried to do too many things and only sort of resolves some of the things

I'm having trouble grokking the specific use cases where users would want to use tools in a throwaway process and not include them in the slug or main process though, so that's context I'm missing

bin/compile
jmorrell: > it results in surprising behavior like environment vars not being shared with the parent process

Ex: If I understand correctly, a script installing a binary can't modify the `PATH` env var in these hooks when they are called with `sub_env` and this is surprising to users

README.md
freeformz: The go-buildpack puts compiled executables in `/bin` and always has (IIRC). This was patterned off `go install`, which puts binaries into `$GOPATH/bin`, so I'm not sure what you mean by you local build puts it in `/`. Do you mean that when you run `go build` the output is placed in `.` ? If so well, the buildpack doesn't use `go build` because `go build` with multiple packages (`./cmd/...`, `./...`, etc) discards the results.

PS: I also generally suggest people not use `go build` locally because of this inconsistency.

--

It's possible that compilation could put an executable named `post-compile` into `/bin`, but IMO that's a feature. i.e. it allows someone to write their post-compile "script" in go, which the buildpack would compile and then execute. So that is the reason I like having it as `/bin/post-compile` and `/bin/pre-compile` for consistency of location. Also, `/bin` is where python keeps them and we shouldn't have 5 different ways to do things.

Which reminds me.... I really prefer `-` over `_`, but python is using `_`, so we should probaby switch to `_` over `-`. Thoughts?



`pre-compile` on the other hand, since it happens before compilation, can't be written in go.

bin/compile
freeformz: bleh. I agree that it could help, but I'm also reluctant to add a chmod. It just seems wrong to blindly chmod. Maybe my graybeard is showing.

bin/compile
freeformz: Okay, then why does the python buildpack use it and why is it part of the buildpack stdlib?

README.md
jmorrell: > This was patterned off `go install`, which puts binaries into `$GOPATH/bin`, so I'm not sure what you mean by you local build puts it in `/`. Do you mean that when you run `go build` the output is placed in `.` ?

This is using go modules if that helps.

> PS: I also generally suggest people not use `go build` locally because of this inconsistency.

I believe I was going off of https://github.com/golang/go/wiki/Modules#quick-start which uses `go build`

```
$ go build 
$ ./hello

Hello, world.
```

As a beginner it seems odd that when pushed to Heroku the binary is put in a different location. Is this something that has changed with go modules that we should revisit?

> Also, `/bin` is where python keeps them and we shouldn't have 5 different ways to do things.
> Which reminds me.... I really prefer `-` over `_`, but python is using `_`, so we should probably switch to `_` over `-`. Thoughts?

I think we're going to have different ways of doing this per-language regardless since it should conform to the community's conventions. I wouldn't worry so much about being consistent here.

Plus I believe the Python functionality was undocumented, though I do not know why. Perhaps @CaseyFaist does.

> It's possible that compilation could put an executable named `post-compile` into `/bin`, but IMO that's a feature. i.e. it allows someone to write their post-compile "script" in go, which the buildpack would compile and then execute.

I do like this bit

jmorrell: > If so well, the buildpack doesn't use `go build` because `go build` with multiple packages (`./cmd/...`, `./...`, etc) discards the results.

Ah, I think I skipped over this, and need to read more

bin/compile
jmorrell: That was added by a previous maintainer and is only used by the python buildpack afaict. I'm not familiar with the original reasoning, but my understanding is that we will likely remove it since in practice it's caused issues and confusion for users.

README.md
jmorrell: > So that is the reason I like having it as `/bin/post-compile` and `/bin/pre-compile` for consistency of location.

Was talking through this today, and someone brought up that this consistency actually works against us in this case. If for some reason someone were to use both the Go and Python buildpacks (or any other buildpack that used the same location for these hooks), then these hooks would be executed twice.

That's not specifically an issue with this since the path name is different, but having `bin/post-compile` and `bin/post_compile` would be a bit odd. 🤔

freeformz: The behavior of `go build` vs. `go install` doesn't change wrt go modules, or at least it doesn't atm. It may in the future. I suspect `go build` is used in most of the docs because Go is (a) trying to get away from $GOPATH and (b) it's simpler for the simple case.

With that said, I still think that the buildpack will continue to need to use `go install`. I can change the location of where it puts binaries though (`GOBIN=${build}`) but that could possibly break some users as well, so it's not a change I want to `just do`.

Re naming conflicts wrt buildpacks and execution of these scripts happening more than once?

How about `/bin/go-post-compile` and `/bin/go-pre-compile` ?

bin/compile
freeformz: changed to: if -f && chmod +x

bin/compile
```diff
@@ -452,8 +452,18 @@ case "${TOOL}" in
         unset GIT_DIR # unset git dir or it will mess with goinstall
         cd ${build}
         export GOBIN="${build}/bin"
+        if [[ -f bin/go-pre-compile ]]; then
+            step "Running bin/go-pre-compile hook"
+            chmod a+x bin/go-pre-compile
```
jmorrell, line 457: TIL about `chmod a+x` 🤓

freeformz:
PS: Sorry for the late responses, I've been sick for the last 3 days.

jmorrell (approved):
This addresses all of the issues that I had with this approach 👍 Thanks for the changes!

jmorrell:
I think I overall still prefer the idea of using a directive in `go.mod` for this purpose, [link](https://github.com/heroku/heroku-buildpack-go/pull/78#issuecomment-458742094) since it prescribes less about the user's directory structure and requires less work from the user.

But this also solves the problem for people who haven't or can't move to go modules yet, so I'm +1

freeformz:
@jmorrell FWIW: I've only implemented this for go modules and have no intent on "backporting" it to work with other tools. Mostly because I want to kill all those other tools ASAP.

freeformz:
If there is a consensus that my `// +heroku` go.mod hack is better then I can do that instead.

But ...

I don't feel that it's *less* work for the user, in fact I feel that it's the complete opposite: more work for the user. They need to not only create a file somewhere, but also have to know to add the magic comment to `go.mod`, make sure there aren't any typos, etc.

freeformz:
@jmorrell Made a point to me (offline) that the `// +heroku` comment hack doesn't require a seperate file, but could embed random commands.

While that's fine and many tools do that: I'm against that as there are a bunch of hidden implications related to executing whatever is in that string. For instance commands that use shell bits like `&&` imply that a shell is being used to execute the script (which is true for now; but may not always be). I also don't want to have to figure out and fight for all the limits of `// +` comment handling in the `go.mod`/go parser.

PR #290, opened 2019-02-20
Add mcount metrics

sowjumn:
Add metrics for different golang version usage.

### DIFF
--- a/bin/compile
+++ b/bin/compile
@@ -12,6 +12,16 @@ buildpack=$(cd "$(dirname $0)/.." && pwd)
 source "${buildpack}/lib/common.sh"
 ensureInPath "jq-linux64" "${cache}/.jq/bin"
 
+# This is used by the buildpack stdlib for metrics
+BPLOG_PREFIX="buildpack.go"
+
+STDLIB_FILE=$(mktemp -t stdlib.XXXXX)
+
+### Load dependencies
+
+curl --silent --retry 5 --retry-max-time 15 'https://lang-common.s3.amazonaws.com/buildpack-stdlib/v8/stdlib.sh' > "$STDLIB_FILE"
+source "$STDLIB_FILE"
+
 DefaultGoVersion="$(<${DataJSON} jq -r '.Go.DefaultVersion')"
 DepVersion="$(<${DataJSON} jq -r '.Dep.DefaultVersion')"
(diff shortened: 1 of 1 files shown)


bin/compile
```diff
@@ -403,6 +412,8 @@ trap clearGitCredHelper INT TERM EXIT
 
 determineTool
 
+mcount "version.$ver"
```
jmorrell: There are other binaries we might track versions of, like `dep`, etc. How about making this:

```
mcount "version.go.$ver
```

bin/compile
```diff
@@ -403,6 +412,8 @@ trap clearGitCredHelper INT TERM EXIT
...
+
 ver=$(expandVer $ver)
```
jmorrell: if I understand correctly, this may modify the `ver` variable. What if we added the `mcount` inside the `reportVer` function instead?

jmorrell (changes requested):

bin/compile
freeformz: isn't that handled by BPLOG_PREFIX?

freeformz: Oh nevermind. Now I understand. I'm slow. ;-)

bin/compile
freeformz: We should record the initial version and the expanded version IMO.

bin/compile
sowjumn: @freeformz @jmorrell these are excellent suggestions. What do you think of 
BPLOG_PREFIX="buildpack.golang" with mcount "version.go.$ver

bin/compile
sowjumn: @jmorrell @freeformz expandVer modifies the `ver` variable with go${v} . I was thinking that as long as we capture the version we should be fine and hence I placed it there. But lets chat about this.

jmorrell: The issue is that you'll capture both: `1.12` and `go1.12` and count them as separate things when you place it here.

That info is also useful, but it's probably not what you want when capturing "which go version was installed"

bin/compile
jmorrell: I lean towards `BPLOG_PREFIX="buildpack.go"` so that it matches the buildpack name, but either works.

jmorrell (approved):
Thanks for the changes 👍

PR #292, opened 2019-02-26
Add go1.12, remove support for go1.10*

freeformz:


CHANGELOG.md
```diff
@@ -1,6 +1,9 @@
 # Go Buildpack Changelog
 
 ## Unreleased
+* Add go1.12 and default to it when go1.12 is specified.
+* Add go1.12 to the list of supported versions.
+* Remove support for go1.10*
```
tt: Nothing gets removed so is this about deprecation?

data.json
```diff
@@ -1,10 +1,10 @@
 {
   "Go": {
-    "Supported": ["go1.10.8", "go1.11.5"],
+    "Supported": ["go1.11.5", "go1.12"],
```
danp, line 3: @tt the removal of support for 1.10 happens here

danp (approved):

data.json
danp: Sorry, think now I misunderstood your question. This "ends official support" for 1.10 but the buildpack will still work with it, so maybe deprecation is a better word.

PR #293, opened 2019-02-28
Version support & metrics

sowjumn:
## Reasons for making these changes
The current buildpack warns with a huge warning even for releases that are from two months ago. We used to support the last two minor versions. We discussed and we would like to support the last two major versions similar to what golang does here https://golang.org/doc/devel/release.html#policy 

## Changes
Now supports two major versions since the last release as per golang
This change also eliminates the 'unsupported' warning for latest releases from like two months ago
Adds metrics to count dependency management systems

### DIFF
--- a/data.json
+++ b/data.json
@@ -1,6 +1,6 @@
 {
   "Go": {
-    "Supported": ["go1.11.5", "go1.12"],
+    "Supported": ["go1.12","go1.11.5","go1.11.4", "go1.11.3", "go1.11.2", "go1.11.1"],
     "DefaultVersion": "go1.11.5",
     "VersionExpansion": {
       "go1.12.0": "go1.12",
@@ -78,7 +78,12 @@
   "test": {
     "assets": [
       "jq-linux64",
+      "go1.12.linux-amd64.tar.gz",
       "go1.11.5.linux-amd64.tar.gz",
+      "go1.11.4.linux-amd64.tar.gz",
+      "go1.11.3.linux-amd64.tar.gz",
+      "go1.11.2.linux-amd64.tar.gz",
+      "go1.11.1.linux-amd64.tar.gz",
       "go1.10.8.linux-amd64.tar.gz",
       "go1.8.3.linux-amd64.tar.gz",
       "go1.7.6.linux-amd64.tar.gz",
--- a/bin/compile
+++ b/bin/compile
@@ -445,6 +445,7 @@ fi
 
 export GOPATH
 
+mcount "pkgmanagement.$TOOL"
 # GB installation
 case "${TOOL}" in
     gomodules)


data.json
```diff
@@ -78,7 +78,12 @@
   "test": {
     "assets": [
       "jq-linux64",
+      "go1.12.linux-amd64.tar.gz",
       "go1.11.5.linux-amd64.tar.gz",
+      "go1.11.4.linux-amd64.tar.gz",
```
jmorrell, line 83: What tests are using these?

jmorrell (approved):

data.json
```diff
@@ -1,6 +1,6 @@
 {
   "Go": {
-    "Supported": ["go1.11.5", "go1.12"],
+    "Supported": ["go1.12","go1.11.5","go1.11.4", "go1.11.3", "go1.11.2", "go1.11.1"],
```
freeformz: `go1.11` needs to be added here if we're supporting all minor versions of the last two major releases.

freeformz, line 3: `go1.11` needs to be added here if we're supporting all minor versions of the last two major releases.

data.json
freeformz: +1 to that question.

freeformz (changes requested):
In addition to my other comment:

I'd like the buildpack to output a warning message if, after expandVer is run, the build isn't being run with the latest minor release of the supported major versions.

I'm not keen on actually *supporting* anything *but* the most recent minor release of the last two major versions. Our discussion in slack, to me at least, was more about changing the error to a warning, not about changing the support policy.

This patch also needs an update to the Changelog.

jmorrell:
> I'd like the buildpack to output a warning message if, after expandVer is run, the build isn't being run with the latest minor release of the supported major versions

See my comment in Slack, but I think we should be defaulting to removing any bit of potential friction for users for now. We can make sure our support policy in the DevCenter to reflect which versions we support but don't need to add flashing lights in the build logs.

freeformz:
@jmorrell Go releases minor versions for *security* and *correctness*, so I think it's appropriate to continue warning users.

freeformz:
I don't understand the `potential friction` comment. Users have to go out of their way to pin themselves to a specific minor version of Go. They'll only get the current warning if they do so.

freeformz:
My take aways from our slack discussion, were to soften the warning from:
```
Deprecated or unsupported version of go ...
```

To something that doesn't reference `unsupported` for the minor versions of Go that are supported.

freeformz:
PS: If there is a doc or example app that specifies a minor version that should be fixed to specify a major release instead.

freeformz (approved):

sowjumn:
Closing this PR for now without merging after the feedback.

freeformz:
I don't understand the `potential friction` comment. Users have to go out of their way to specify a minor version. All docs/examples should generally be guiding people to using `go1.X`, not `go1.X.Y`. The go team releases minor versions to address **security and correctness** issues. New functionality is **not** added in minor releases.

With that said: I've decided that my objections should be noted, but shouldn't hold up what you are trying to do.

PR #296, opened 2019-03-01
Removed too many warnings

sowjumn:
Removed too many warnings

### DIFF
--- a/bin/compile
+++ b/bin/compile
@@ -642,7 +642,9 @@ if [ "${TOOL}" != "gb" ]; then
     echo "NAME=${name}" >> "${t}"
 fi
 
-info ""
-info "We placed the compiled binary in ./bin on your dyno."
-info "If you want to invoke it in your Procfile use ./bin/your-installed-directory-name"
-info ""
+bin_path="${build}/bin"
+bin_files=`ls ${bin_path}`
+echo -e "\nCompiled the following binaries:"
+for fyle in $bin_files; do
+  info "$(echo ./bin/$fyle)\n"
+done 
--- a/bin/compile
+++ b/bin/compile
@@ -646,5 +646,6 @@ bin_path="${build}/bin"
 bin_files=`ls ${bin_path}`
 echo -e "\nCompiled the following binaries:"
 for fyle in $bin_files; do
-  info "$(echo ./bin/$fyle)\n"
+  info "$(echo ./bin/$fyle)"
+  echo -e "\n"
 done 
--- a/bin/compile
+++ b/bin/compile
@@ -648,5 +648,4 @@ echo -e "\nCompiled the following binaries:"
 for fyle in $bin_files; do
   info "$(echo ./bin/$fyle)"
 done 
-
-echo -e "\n"
+echo ""
--- a/bin/compile
+++ b/bin/compile
@@ -647,5 +647,4 @@ bin_files=`ls ${bin_path}`
 echo -e "\nCompiled the following binaries:"
 for fyle in $bin_files; do
   info "$(echo ./bin/$fyle)"
-  echo -e "\n"
 done 
--- a/bin/compile
+++ b/bin/compile
@@ -648,3 +648,5 @@ echo -e "\nCompiled the following binaries:"
 for fyle in $bin_files; do
   info "$(echo ./bin/$fyle)"
 done 
+
+echo -e "\n"
--- a/bin/compile
+++ b/bin/compile
@@ -47,8 +47,6 @@ handleDefaultPkgSpec() {
                 warn "To install a different package spec add a comment in the following form to your \`go.mod\` file:"
                 warn "// +heroku install ./cmd/..."
                 warn ""
-                warn "For more details see: https://devcenter.heroku.com/articles/go-apps-with-modules#build-configuration"
-                warn ""
             ;;
             dep)
                 warn "To install a different package spec set 'metadata.heroku.install' in 'Gopkg.toml'"
@@ -115,15 +113,11 @@ expandVer() {
     echo $(<${DataJSON} jq -r 'if .Go.VersionExpansion."'${v}'" then .Go.VersionExpansion."'${v}'" else "'${v}'" end')
 }
 
-# Report deprecated versions to user
+# Report development versions to user
 # Use after expandVer
 reportVer() {
     local ver="${1}"
     mcount "version.go.$ver"
-
-    if <"${DataJSON}" jq  -e '.Go.Supported | any(. == "'${ver}'")' &> /dev/null; then
-        return
-    fi
 
     case $ver in
         devel*)
@@ -133,12 +127,6 @@ reportVer() {
             warn "but is otherwise unsupported."
             warn ""
             warn "Build tests are NOT RUN!!"
-            warn ""
-        ;;
-        *)
-            warn ""
-            warn "Deprecated or unsupported version of go ($ver)"
-            warn "See https://devcenter.heroku.com/articles/go-support#go-versions for supported version information."
             warn ""
         ;;
     esac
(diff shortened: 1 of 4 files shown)


bin/compile
```diff
@@ -651,3 +639,6 @@ echo "TOOL=${TOOL}" > "${t}"
 if [ "${TOOL}" != "gb" ]; then
     echo "NAME=${name}" >> "${t}"
 fi
+
+warn "We placed the compiled binary in /bin on your dyno."
```
jmorrell: Is `warn` the correct command here? Do you intend for this to be yellow with exclamation marks? If you want it to be regulary text, you can use `echo`, otherwise 👍 

Could you attach a screenshot of this message in context?

lib/common.sh
```diff
@@ -274,12 +274,7 @@ determineTool() {
     if [ -f "${goMOD}" ]; then
         TOOL="gomodules"
         warn ""
-        warn "Go modules are an experimental feature of go1.11+"
-        warn "Any issues building code that uses Go modules should be"
-        warn "reported via: https://github.com/heroku/heroku-buildpack-go/issues"
-        warn ""
-        warn "Additional documentation for using Go modules with this buildpack"
-        warn "can be found here: https://github.com/heroku/heroku-buildpack-go#go-module-specifics"
+        warn "Found a go.mod file and using Go modules"
```
jmorrell: I don't think we want the `!!!` in the message here. I would suggest something like:

```
step "Detected go modules - go.mod"
```

and when we drop support for other package managers we can delete this line

bin/compile
sowjumn: echo sounds great!

lib/common.sh
sowjumn: Excellent. Yes will do this.

jmorrell:
Also, can you update the CHANGELOG alongside these changes?

bin/compile
```diff
@@ -651,3 +639,8 @@ echo "TOOL=${TOOL}" > "${t}"
 if [ "${TOOL}" != "gb" ]; then
     echo "NAME=${name}" >> "${t}"
 fi
+
+info ""
+info "We placed the compiled binary in ./bin on your dyno."
+info "If you want to invoke it in your Procfile use ./bin/your-installed-directory-name"
```
jmorrell: It would be more work, but something customized to the users app like this might be nicer:

```
Compiled binaries:
   ./bin/web
   ./bin/other-name
   ./bin/etc
```

lib/common.sh
```diff
@@ -294,7 +294,7 @@ determineTool() {
             warn ""
         fi
         if ! <"${DataJSON}" jq  -e '.Go.SupportsModuleExperiment | any(. == "'${ver}'")' &> /dev/null; then
-            err "You are using ${ver}, which does not support the Go modules experiment"
+            err "You are using ${ver}, which does not support the Go modules"
```
bernerdschaefer: Nitpick:

```suggestion
            err "You are using ${ver}, which does not support Go modules"
```

bin/compile
```diff
@@ -47,8 +47,6 @@ handleDefaultPkgSpec() {
                 warn "To install a different package spec add a comment in the following form to your \`go.mod\` file:"
                 warn "// +heroku install ./cmd/..."
                 warn ""
-                warn "For more details see: https://devcenter.heroku.com/articles/go-apps-with-modules#build-configuration"
```
bernerdschaefer, line 50: I was going to say that this seemed useful, because there are lots of build details to customize to get things really working well for Go.. but the decventer article is empty. So 👍 at least until we've got better docs in place.

bernerdschaefer (approved):

bin/compile
```diff
@@ -651,3 +639,8 @@ echo "TOOL=${TOOL}" > "${t}"
 if [ "${TOOL}" != "gb" ]; then
     echo "NAME=${name}" >> "${t}"
 fi
+
+info ""
+info "We placed the compiled binary in ./bin on your dyno."
```
freeformz: Some thoughts:

- IMO we should not use personal pronouns in messages to customers. "We" didn't place anything anywhere. The software we wrote, this buildpack, did.

- The buildpack may build more than one binary.

- The buildpack is not building a dyno, it's compiling the customer's code.

How about something like this:
```
Compiled the following binaries:
\t$HOME/bin/foo
\t$HOME/bin/bar
\t$HOME/bin/bing
```

The buildpack would need to be modified to determine what was installed, the easiest way to do that is to probably list the contents of `$build/bin` before anything is installed (code, tools, etc) and then afterwards, removing anything from the after list that was also in the before list. 

This can be done with something like this:
```sh
_oifs=$IFS
IFS=$'\n'
before=($(shasum ${build}/bin/*))
IFS=$_oifs

# ... Install the things

_oifs=$IFS
IFS=$'\n'
after=($(shasum ${build}/bin/*))
IFS=$_oifs

# Determine what is new
new=()
for a in "${after[@]}"; do
  let found=0

  for b in "${before[@]}"; do
    if [ "${a}" == "${b}" ]; then
      let found+=1
    fi
  done

  if [ $found -eq 0 ]; then
    new+=( $(echo $a | awk '{print $2}' ) )
  fi
done

echo "Compiled the following binaries:"
for n in "${new[@]}"; do
  echo -e "\t$(echo $n| sed 's/^\./\$HOME/')"
done
```

bin/compile
freeformz: The go buildpack adds $HOME/bin to PATH, so it's not necessary to include the full path. Additionally there is no guarantee that when a dyno is booting that by the time the Procfile entry is run that the current working directory will be /app, so I have stayed away from using `./bin/` as a prefix in the Procfile.

bin/compile
```diff
@@ -135,12 +129,6 @@ reportVer() {
             warn "Build tests are NOT RUN!!"
             warn ""
         ;;
-        *)
-            warn ""
-            warn "Deprecated or unsupported version of go ($ver)"
```
freeformz, line 140: FWIW: As a user, if I have pinned to what is now an older, unsupported version of Go, I want to be told that my version of GO is no longer supported/deprecated as I may not have time to keep track of go releases.

bin/compile
freeformz: Yeah, my intent was to write that article. Reviews at the time, or previously (I forget which), asked for a blank article link at least.

lib/common.sh
freeformz: Not for this PR, but as a follow on one it may be better to list the version of Go that don't support modules. This is similar to how vendoring landed as well. At first the buildpack listed the versions that supported it, then it was switched so that it listed the versions that didn't.

freeformz (changes requested):

freeformz (approved):

bin/compile
```diff
@@ -115,16 +113,12 @@ expandVer() {
...
 }
 
-# Report deprecated versions to user
+# Report development versions to user
 # Use after expandVer
 reportVer() {
     local ver="${1}"
     mcount "version.go.$ver"
 
-    if <"${DataJSON}" jq  -e '.Go.Supported | any(. == "'${ver}'")' &> /dev/null; then
```
freeformz, line 124: Maybe we change this to supported major versions instead? I think it's worth telling people who are still running anything older than go1.11 that they are building using an unsupported version of Go.

bin/compile
sowjumn: Last I heard this is not how other languages work. Our support policy is/will be listed on the heroku site and I would default to as little warnings as I can.

lib/common.sh
sowjumn: Sounds good! I took notes to add this

bin/compile
sowjumn: I will add this as a separate feature.

bin/compile
sowjumn: resolving this in the intent of adding warning for supported major versions

bin/compile
sowjumn: Excellent suggestion. I will do that

PR #300, opened 2019-03-08
Add codeowners for code review

jmorrell:
This brings the go buildpack into line with the rest of the official buildpacks.

Though I believe we either need to add a few more people to the @heroku/languages team, or expand this file.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,2 @@
+* @heroku/languages
+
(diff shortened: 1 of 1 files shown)


freeformz (approved):
If it's okay with @heroku/languages I'd like to be added to this file as well, at least for now.

jmorrell:
@freeformz mostly I was unsure how to do that and thought adding you to the team might be simpler. I'll take a stab at what I think the docs say here

PR #302, opened 2019-03-12
Fix up gomodules testing a little bit

freeformz:
If GOPATH is unset, the $HOME/go is the default. Which can cause
the benchmark run to also test any downloaded dependencies.

If `./vendor` exists add -mod=vendor to pin to the contents of vendor.

If not, then, as per the go docs, add -mod=readonly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Go Buildpack Changelog
 
 ## Unreleased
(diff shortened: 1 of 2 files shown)


freeformz:
Thanks to @sheax0r for patiently pointing out the problems to me.

sheax0r (approved):

PR #304, opened 2019-03-12
Modules install

sowjumn:
* Refactored pkgs install 
* Buildpack now checks for cmd directory and installs it automatically 
* If Buildpack doesn't find cmd directory install project root directory packages

### DIFF
--- a/bin/compile
+++ b/bin/compile
@@ -395,6 +395,11 @@ setupGOPATH() {
     echo "src=${src}"
 }
 
+installPkgs() {
+    step "Running: go install -v ${FLAGS[@]} ${pkgs}"
+    go install -v "${FLAGS[@]}" ${pkgs} 2>&1
+}
+
 loadEnvDir "${env_dir}"
 
 setGitCredHelper "${env_dir}"
@@ -440,7 +445,15 @@ case "${TOOL}" in
     gomodules)
         pkgs=${GO_INSTALL_PACKAGE_SPEC:-$(awk '{ if ($1 == "//" && $2 == "+heroku" && $3 == "install" ) { print substr($0, index($0,$4)); exit }}' ${goMOD})}
         if [ -z "${pkgs}" ]; then
-            pkgs="default"
+            if [ -d "${build}/cmd" ]; then
+                # Supports standard Go directory structure  https://github.com/golang-standards/project-layout#go-directories
+                step "Detected \"cmd\" directory. Installing packages under \"cmd\""
+                mcount "pkginstall.cmd"
+                pkgs="./cmd/..."
+            else
+                mcount "pkginstall.default"
+                pkgs="default"
+            fi
         fi
         warnPackageSpecOverride
         handleDefaultPkgSpec
@@ -458,8 +471,9 @@ case "${TOOL}" in
             chmod a+x bin/go-pre-compile
             bin/go-pre-compile
         fi
-        step "Running: go install -v ${FLAGS[@]} ${pkgs}"
-        go install -v "${FLAGS[@]}" ${pkgs} 2>&1
+
+        installPkgs
+
         if [[ -f bin/go-post-compile ]]; then
             step "Running bin/go-post-compile hook"
             chmod a+x bin/go-post-compile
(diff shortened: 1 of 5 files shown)


bin/compile
```diff
@@ -395,6 +395,11 @@ setupGOPATH() {
     echo "src=${src}"
 }
 
+installPkgs() {
+	step "Running: go install -v ${FLAGS[@]} ${pkgs}"
```
jmorrell: nit: indentation

bin/compile
```diff
@@ -440,7 +445,14 @@ case "${TOOL}" in
     gomodules)
         pkgs=${GO_INSTALL_PACKAGE_SPEC:-$(awk '{ if ($1 == "//" && $2 == "+heroku" && $3 == "install" ) { print substr($0, index($0,$4)); exit }}' ${goMOD})}
         if [ -z "${pkgs}" ]; then
-            pkgs="default"
+            if [ -d "${build}/cmd" ]; then
+                step "Detected \`cmd\` folder. Installing packages under \`./cmd\`"
+                step "To install packages under MyDir/ add the following to \`go.mod\` file:"
+                warn "// +heroku install ./MyDir/..."
+                pkgs="./cmd/..."
```
jmorrell, line 452: I would suggest adding a comment with a link to https://github.com/golang-standards/project-layout#go-directories

bin/compile
```diff
@@ -440,7 +445,14 @@ case "${TOOL}" in
     gomodules)
         pkgs=${GO_INSTALL_PACKAGE_SPEC:-$(awk '{ if ($1 == "//" && $2 == "+heroku" && $3 == "install" ) { print substr($0, index($0,$4)); exit }}' ${goMOD})}
         if [ -z "${pkgs}" ]; then
-            pkgs="default"
+            if [ -d "${build}/cmd" ]; then
```
jmorrell, line 448: Can we add `mcount`s to both of these branches so we can see how often each is used?

bin/compile
```diff
@@ -440,7 +445,14 @@ case "${TOOL}" in
...
+                step "Detected \`cmd\` folder. Installing packages under \`./cmd\`"
+                step "To install packages under MyDir/ add the following to \`go.mod\` file:"
```
jmorrell: > "Detected \`cmd\` folder. Installing packages under \`./cmd\`"

- The backticks are unnecessary when viewed in the terminal
- I would prefer to not have the `./` as not everyone may be familiar with that unix convention. Perhaps we could list the different "applications" we're installing dependencies for?

Ex:

```
Detected "cmd" folder. Installing packages for:
  cmd/web
  cmd/migrate
  cmd/worker
```

jmorrell: > step "To install packages under MyDir/ add the following to \`go.mod\` file:"
> warn "// +heroku install ./MyDir/..."

Do we need this warning? It will be displayed for every build

bin/compile
sowjumn: Yes I will

bin/compile
sowjumn: lets chat about this.

bin/compile
sowjumn: I will add this: Supports standard go directory structure with the link

bin/compile
freeformz: `./cmd/...` is a go thing, inspired by the unix convention, so if you're writting go, you'll run into it and likely be familiar with it.

freeformz: PS: Don't confuse a package spec with a directory, they're not the same thing, even though packages are contained within directories.

For instance, given a package named `github.com/foo/bar`, if the buildpack were building that code and it contained the `cmd/foo` and `cmd/bar` directories, both with main packages in them, then the package spec of `./cmd/...` and `github.com/foo/bar/cmd/...` are essentially equivalent.

freeformz: With that said, it would be nice to detail what package spec we've decided AND what commands we're installing because of it.

freeformz (dismissed):
If this had a test or tests to verify the change I would approve it.

Everything else is a nit or can be addressed in follow up PRs.

freeformz (approved):

PR #309, opened 2019-03-28
[Go Modules] Detect main packages and install those when no package spec is specified

### LINKED ISSUE #308: v105 breaks root-level package installation if cmd dir exists
PR #304 broke the installation of root-level packages if a `cmd` directory exists. Unless I've missed something, `cmd` is not an officially reserved directory name for go modules/packages. The only reference to it I was able to find was in the **unofficial** guide: https://github.com/golang-standards/project-layout#go-directories

While I can appreciate that layout, it is not official and therefore I think it's incorrect to push a breaking update, at least without any notice (e.g. there could have been an echo in the script saying something like "AS OF v105, ROOT LEVEL PACKAGES WILL NOT BE INSTALLED IF A cmd DIRECTORY EXISTS, PLEASE SEE <url> FOR MORE INFO").

One have to keep in mind that there might be older/bigger applications that happen to have a cmd package, but that still utilizes the root package as the main one (e.g. the one to build a binary from). Such is the case for me at the moment. Took a bit of debugging to figure out why the main binary stopped building all of a sudden (the cryptic "detected cmd directory" comment didn't help much).

Nonetheless, I followed the instructions that I saw somewhere that said to add the following comment to the go.mod: `// +heroku install ./...`. That restored the behavior.

Anyway, if you're planning on keeping this change as-is, a more clear notice should be added to the script (towards the top) imo.

freeformz:
Closes #308 

This PR does a few things:
1. For go modules, detect *any* main packages in the repo, that are not in vendor, and when there isn't a specified package spec, install those packages.
1. Instead of listing all of bin/* as being what the buildpack has installed, take an shasum snapshot of the contents of bin/ (if it exists) before doing anything, and a snapshot at the end and detail what files the buildpack has installed.
1. Small set of related(ish) README updates.

### DIFF
--- a/bin/compile
+++ b/bin/compile
@@ -444,6 +444,7 @@ export GOPATH
 mcount "pkgmanagement.$TOOL"
 
 mainPackagesInModule() {
+  # For an explanation of what this is doing, see https://dave.cheney.net/2014/09/14/go-list-your-swiss-army-knife
   go list -find -mod=vendor -f '{{ if eq .Name "main" }} {{.ImportPath}} {{ end }}' ./...
 }
 
@@ -461,7 +462,7 @@ case "${TOOL}" in
                 warn "Unsure what package(s) to install, defaulting to '.'"
                 warn ""
                 pkgs="default"
-                mcount "pkginstall.default"
+                mcount "pkginstall.detect-failed"
             else
                 info ""
                 info "Detected the following main packages to install:"
@@ -480,7 +481,7 @@ case "${TOOL}" in
         fi
 
         if [[ "${pkgs}" =~ "..." ]]; then
-            mcount "pkginstall.dot_dot_dot"
+            mcount "package_spec.contains.dot_dot_dot"
         fi
 
         warnPackageSpecOverride
--- a/bin/compile
+++ b/bin/compile
@@ -657,6 +657,6 @@ bin_path="${build}/bin"
 bin_files=`ls ${bin_path}`
 echo -e "\nCompiled the following binaries:"
 for fyle in $bin_files; do
-  info "$(echo ./bin/$fyle)"
+  info "./bin/$fyle"
 done 
 echo ""
--- a/bin/compile
+++ b/bin/compile
@@ -15,12 +15,12 @@ ensureInPath "jq-linux64" "${cache}/.jq/bin"
 # This is used by the buildpack stdlib for metrics
 BPLOG_PREFIX="buildpack.go"
 
-STDLIB_FILE=$(mktemp -t stdlib.XXXXX)
+STDLIB_DIR=$(mktemp -d -t stdlib.XXXXX)
 
 ### Load dependencies
 
-curl --silent --retry 5 --retry-max-time 15 'https://lang-common.s3.amazonaws.com/buildpack-stdlib/v8/stdlib.sh' > "$STDLIB_FILE"
-source "$STDLIB_FILE"
+ensureFile "stdlib.sh.v8" "${STDLIB_DIR}" "chmod a+x"
+source "${STDLIB_DIR}/stdlib.sh.v8"
 
 DefaultGoVersion="$(<${DataJSON} jq -r '.Go.DefaultVersion')"
 DepVersion="$(<${DataJSON} jq -r '.Dep.DefaultVersion')"
--- a/bin/compile
+++ b/bin/compile
@@ -443,6 +443,8 @@ mcount "pkgmanagement.$TOOL"
 
 case "${TOOL}" in
     gomodules)
+        cd ${build}
+        step "Determining packages to install"
         pkgs=${GO_INSTALL_PACKAGE_SPEC:-$(awk '{ if ($1 == "//" && $2 == "+heroku" && $3 == "install" ) { print substr($0, index($0,$4)); exit }}' ${goMOD})}
         if [ -z "${pkgs}" ]; then
             if [ -d "${build}/cmd" ]; then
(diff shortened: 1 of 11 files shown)


files.json
```diff
@@ -415,6 +415,10 @@
     "SHA": "cc4b544f982fb3cf0ac84fe9c78f3fc37513f31daba90f247e9d477f5daf2396",
     "URL": "https://github.com/golang-migrate/migrate/releases/download/v3.4.0/migrate.linux-amd64.tar.gz"
   },
+  "stdlib.sh.v8": {
+    "SHA": "48b1c663ceda9e93a42fa2ed610fbdd12f9801abd55c317990b30ec75dc3d715",
```
jmorrell, line 419: When trying to deploy I got this:

```
remote: -----> Go app detected
remote: -----> Fetching jq... done
remote: -----> Fetching stdlib.sh.v8...  !!    
remote:  !!    Downloaded file (stdlib.sh.v8) sha does not match recorded SHA
remote:  !!    Unable to continue.
remote:  !!    
remote:  !     Push rejected, failed to compile Go app.
```

but I'm not sure how CI would pass if this is the case?

bin/compile
```diff
@@ -10,17 +10,19 @@ env_dir="${3}"
...
 BPLOG_PREFIX="buildpack.go"
 
-STDLIB_FILE=$(mktemp -t stdlib.XXXXX)
+STDLIB_DIR=$(mktemp -d -t stdlib.XXXXX)
 
 ### Load dependencies
 
-curl --silent --retry 5 --retry-max-time 15 'https://lang-common.s3.amazonaws.com/buildpack-stdlib/v8/stdlib.sh' > "$STDLIB_FILE"
-source "$STDLIB_FILE"
+ensureFile "stdlib.sh.v8" "${STDLIB_DIR}" "chmod a+x"
```
jmorrell, line 24: this logs out to the console:

```
remote: -----> Fetching stdlib.sh.v8... done
```

I don't believe this information isn't helpful to the user. They don't know what a `stdlib.sh.v8` is, or why it would need to be part of their build process. I think we should only be logging something out if this download fails, which will be exceedingly rare.

bin/compile
```diff
@@ -441,20 +443,46 @@ export GOPATH
 # GB installation
 mcount "pkgmanagement.$TOOL"
 
+mainPackagesInModule() {
+  go list -find -mod=vendor -f '{{ if eq .Name "main" }} {{.ImportPath}} {{ end }}' ./...
```
jmorrell, line 448: Well that is nifty 😮 

This is a bit magic though. Can you add some comments about how this works for anyone who comes across it? Maybe a link to https://dave.cheney.net/2014/09/14/go-list-your-swiss-army-knife ?

bin/compile
```diff
@@ -441,20 +443,46 @@ export GOPATH
...
-                mcount "pkginstall.default"
+            pkgs=$(mainPackagesInModule)
+            if [ -z "${pkgs}" ]; then
+                warn ""
+                warn "No main packages found in module ${name}"
+                warn ""
+                warn "Unsure what package(s) to install, defaulting to '.'"
+                warn ""
                 pkgs="default"
+                mcount "pkginstall.default"
```
jmorrell: suggested: `mcount "pkginstall.detect-failed"`

bin/compile
```diff
@@ -441,20 +443,46 @@ export GOPATH
...
+                    info "\t\t${pkg}"
+                done
+                info ""
+                mcount "pkginstall.auto"
+            fi
+        else
+            if [ -z "${GO_INSTALL_PACKAGE_SPEC}" ]; then
+                mcount "pkginstall.go_install_package_spec"
+            else
+                mcount "pkginstall.heroku_install"
```
jmorrell, line 479: Under what conditions would this branch be hit?

bin/compile
```diff
@@ -441,20 +443,46 @@ export GOPATH
...
             fi
         fi
+
+        if [[ "${pkgs}" =~ "..." ]]; then
+            mcount "pkginstall.dot_dot_dot"
```
jmorrell: Is the purpose here to see how many users use the `...` to define their package spec?

If so, I would suggest using a different namespace for the `mcount` here, so that only one `pkginstall.*` is logged per build.

```
mcount "dot_dot_dot_package_spec"
```

jmorrell (changes requested):
Some minor questions, and concern that trying to deploy with the branch failed for me

bin/compile
freeformz: This is using the standard download and verify routines that everything else uses, so I'd really like to keep it and not have to continue special casing it.

files.json
freeformz: Is it possible someone pushed a new v8 to the bucket?

bin/compile
jmorrell: That's fair

bin/compile
freeformz: When a customer specified exactly what package spec(s) they want to install.

files.json
freeformz: Yeah, that doesn't make sense. Did it work if you pushed again? Perhaps there was an error downloading the file that was squashed/obfuscated by the functions that download and verify files?

freeformz: Esp since I re-verified the file and the recorded sha is correct.

freeformz:
@jmonster I've addressed your feedback, please re-review.

files.json
jmorrell: No, that hasn't been touched in a long time...

I manually verified the SHA and tried deploying again and it worked, so I'm not sure what was happening here.

freeformz:
@jmorrell I've addressed your feedback, please re-review.

jmorrell (approved):

PR #317, opened 2019-04-06
Add go1.12.2 / go1.11.7

freeformz:
Drop support for `Go.SupportsModuleExperiment`. It's not really
and expermiment anymore and any version of go >= go1.11 will be
supporting it.

Drop support for `Go.Supported` since the buildpack isn't using it
anymore. :-(

### DIFF
--- a/lib/common.sh
+++ b/lib/common.sh
@@ -318,6 +318,12 @@ setGoVersionFromEnvironment() {
     ver=${GOVERSION:-$DefaultGoVersion}
 }
 
+supportsGoModules() {
+    local version="${1}"
+    # Ex:      "go1.10.4" | ["go1","10", "4"] | ["1","10","4"]     | [1,10,4]      |  [1]           [10]      == exit 1 (fail)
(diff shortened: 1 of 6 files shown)


lib/common.sh
```diff
@@ -338,16 +338,16 @@ determineTool() {
             warn "For more details see: https://devcenter.heroku.com/articles/go-apps-with-modules#build-configuration"
             warn ""
         fi
-        if ! <"${DataJSON}" jq  -e '.Go.SupportsModuleExperiment | any(. == "'${ver}'")' &> /dev/null; then
+        if ! echo "\"${ver}\"" | jq -e 'split(".") | map(gsub("go";"")) | map(tonumber) | .[0] >= 1 and .[1] >= 11' &> /dev/null; then
```
jmorrell: A comment would be very helpful here. This line is very difficult to parse

```
# if using 1.11 or older
```

`.[0] >= 1 and .[1] >= 11` is particularly confusing because it looks like it's testing whether that is true instead of catching when it fails. Wrapping this into a `goVersionSupportsModules` function might make this easier for others.

jmorrell (approved):

lib/common.sh
freeformz: done.

PR #324, opened 2019-04-25
Setup Procfile for apps using Go Modules when no Procfile exists

freeformz:
When no Procfile exists and

* only a single main package exists, setup the resulting executable as
  the web process type. This maintains existing behavior.

* multiple main packages exist, setup the resulting executables as
  process types of the same name. This means that a main package in a
  `web` directory will be setup as the web process type, a main package
  in a `worker` directory will be setup as the worker process type, etc.

This should make it very simple to push a go app and have Heroku
generally DTRT wrt process types.

This PR also tells the user what Procfile was generated and tells them
to create a Procfile if the generated one doesn't suit their needs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,12 @@
 # Go Buildpack Changelog
 
 ## Unreleased
+* *GoModules* When no Procfile exists and only a single main package exists, setup the resulting executable as the web process type.
+* *GoModules* When no Procfile exists and multiple main packages exist, setup the resulting executables as process types of the same name.
+* *GoModules* This means that a main package in a `web` directory will be setup as the web process type, a package in a `worker` directory will be setup as the worker process type, etc.
 
 ## v111 (2019-04-18)
-* Set GOPATH when using Go modules to capture downloaded dependencies.
+* *GoModules* Set GOPATH to capture downloaded dependencies.
 
 ## v110 (2019-04-15)
 * Add go1.12.4, expand go1.12 to go1.12.4, and default to go1.12.4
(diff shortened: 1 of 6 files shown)


jkutner:
@freeformz is there a reason not to do this in `bin/release` (there may be, i'm only curious. sometimes you need the context from `bin/compile`).

freeformz:
@jkutner The `bin/release` of the buildpack has always been a [noop](https://github.com/heroku/heroku-buildpack-go/commits/master/bin/release).

And yes, until we drop support for everything except Go Modules I feel that doing it in `bin/compile` is easier than `bin/release`.

With that said, I can add a commit that does it there if you can explain or point me to the value of doing it there.

jkutner:
@freeformz i think the value would be the ability to add the `web` process type even if there is a `Procfile`. For example, some one could define `worker: echo "foo"`, and you'd still get to add the `web`. But that may not be that valuable, idk.

jkutner (approved):

sowjumn (approved):

freeformz:
@jkutner The details around [`bin/release`](https://devcenter.heroku.com/articles/buildpack-api#bin-release) are few. Do the contents emitted by `/bin/release` take precedence over the Procfile, vice versa, or are they merged with the Procfile? If they're merged, then which one takes precedence? Is that logic applied to every build? Or just the first build? etc.

If the contents of a user supplied Procfile are merged with the output of /bin/release and the user supplied Procfile takes precedence then I'm more than happy to move all of this logic to `/bin/release`.

jkutner:
@freeformz i don't think there's a right or wrong between bin/release and what you're doing. It really depends on what experience you want. Once you drop Go Modules, it might make sense to move it into bin/release, but i think this is fine for now.

freeformz:
@jkutner The move to `/bin/release` may be an overall improvement depending on the exact details about how `bin/release` and Procfiles interoperate.

PR #326, opened 2019-05-01
Run golanglint-ci when a golanglint-ci config file exists

freeformz:
Many teams generally include some sort of linting pass in their CI configs. This is basically one tool to rule them all with a nice sentinel config file that can be used to detect intent.

cc @meekmichael

### DIFF
--- a/bin/test
+++ b/bin/test
@@ -3,11 +3,11 @@
 
 set -eo pipefail
 
-mkdir -p "${1}" "${2}" # "${3}"
+mkdir -p "${1}" "${2}"
 build="$(cd ${1}/ && pwd)"
 env_dir="$(cd ${2}/ && pwd)"
-#artifact_dir="$(cd ${3} && pwd)"
 testpack=$(cd "$(dirname $0)/.." && pwd)
+buildpack="${testpack}"
 source "${testpack}/lib/common.sh"
 
 # loadEnvDir "${env_dir}"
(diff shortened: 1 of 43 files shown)


bin/test
```diff
@@ -45,8 +45,12 @@ output=$(mktemp)
 case "${TOOL}" in
     gomodules)
         cd "${build}"
+        if [ -f "${build}/.golangci.yml" -o -f "${build}/.golangci.toml" -o -f "${build}/.golangci.json" ]; then
```
jmorrell, line 53: Could you add an `mcount` to track usage here?

lib/common.sh
```diff
@@ -4,6 +4,9 @@
 # load environment variables
 # allow apps to specify cgo flags. The literal text '${build_dir}' is substituted for the build directory
 
+test -n "${_COMMON_LOADED:-}" && exit 0
+export _COMMON_LOADED="true"
```
jmorrell: This is a nifty bash pattern

jmorrell (approved):

bin/test
freeformz: Sure. But, what specifically do you want to track:
* The fact that there is a config file?
* That there isn't a config file?
* That the config file is yaml? Or toml? Or json?
* That golangci-lint was run? Or not run?
* All of the above?
* Something else?

jmorrell: specifically that golangci-lint was run and this feature was used

freeformz: @jmorrell Reminder: this is testpack code. Does the service that execute testpack stuff support the metrics stuff that the build system supports? Last time I checked, which was admittedly a long time ago, it didn't.

jmorrell: ahh, that's a very good point. Could you add a check for these files to `bin/compile` then?

PR #343, opened 2019-06-29
Rewrite sbin/sync-files.sh

freeformz:
### Relace s3cmd with the aws cli.

I've been having problems with s3cmd and I couldn't figure them out.
I've started to use the aws cli in other things, so trying to keep it
standard.

### Parallel shasums

It's been taking longer and longer for them, so kick them all off
and gather up the results via a fifo. This will max out your CPUs, but
that's a good thing.

### Switch to zsh

fifo/signal handling in bash is brain dead and wouldn't work reliably.
Works fine in zsh.

### Add Ability to ignore files

There are times where there is other work in progress that has added
files to the s3 bucket and I don't want to delete them from the bucket.
This option allows the user to ignore those files.

### DIFF
--- a/sbin/sync-files.sh
+++ b/sbin/sync-files.sh
@@ -22,11 +22,11 @@ fi
 
 # assumes we are in bin, find the directory above that
 cwd="$(cd $(dirname $0); cd ..; pwd)"
-jf="${cwd}/files.json"
+filesJSON="${cwd}/files.json"
 
-td="${cwd}/file-cache"
-mkdir -p "${td}"
-cd "${td}"
+cache="${cwd}/file-cache"
+mkdir -p "${cache}"
+cd "${cache}"
 
 echo "Getting bucket credentials"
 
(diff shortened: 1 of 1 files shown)


sbin/sync-files.sh
```diff
@@ -28,52 +30,95 @@ cd "${td}"
...
     u="$(< "${jf}" jq -r '."'${f}'".URL')"
-    echo "Downloading: ${u}"
-    curl -J -o "${f}" -L --retry 15 --retry-delay 2 $u
+    curl -s -J -o "${f}" -L --retry 15 --retry-delay 2 $u 2>&1
   fi
 
-  s="$(< "${jf}" jq -r '."'${f}'".SHA')"
-  if [ ${#s} -eq 40 ]; then
-    sf="$(shasum "${f}" | cut -d \  -f 1)"
+  local sk="$(< "${jf}" jq -r '."'${f}'".SHA' 2>&1)"
```
jmorrell: I find bash kind of cryptic at the best of times, and that combined with the short filenames like `sk`, `jf`, `u` and `sf` here, make it difficult for me to follow the logic

jmorrell (approved):

sbin/sync-files.sh
freeformz: sk == sha known
jf == json file
sf == sha file
u == url

freeformz: But yeah. I can rename them.

jmorrell: none of that was obvious to me 😅

PR #347, opened 2019-07-10
Record the botched releases

freeformz:
v121 errored, so I published again, but that led to v122, so I don't
know what the contents of v121 are.

v120 was my mistake because I forgot to pull.


freeformz:
I added 123-126 here as well.

jmorrell (approved):

PR #361, opened 2019-08-29
add go1.13rc2, use it for go1.13 expansion

danp:
https://groups.google.com/d/msg/golang-announce/JNU1zuzxGug/0ZWnrvplBAAJ

### DIFF
--- a/data.json
+++ b/data.json
@@ -2,8 +2,8 @@
   "Go": {
     "DefaultVersion": "go1.12.9",
     "VersionExpansion": {
-      "go1.13": "go1.13rc1",
-      "go1.13.0": "go1.13rc1",
+      "go1.13": "go1.13rc2",
+      "go1.13.0": "go1.13rc2",
       "go1.12.0": "go1.12",
       "go1.12": "go1.12.9",
       "go1.11.0": "go1.11",
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,9 @@
 # Go Buildpack Changelog
 
 ## Unreleased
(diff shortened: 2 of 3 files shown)


freeformz (approved):
I also sync'd the files.

amerine (approved):

data.json
```diff
@@ -2,8 +2,8 @@
   "Go": {
     "DefaultVersion": "go1.12.9",
     "VersionExpansion": {
-      "go1.13": "go1.13rc1",
-      "go1.13.0": "go1.13rc1",
+      "go1.13": "go1.13rc2",
+      "go1.13.0": "go1.13rc2",
```
jkutner, line 6: It seems a little dangerous to advertise an RC or beta as `1.13`. For other languages we don't expand these, and require the full version to be specified. This prevents someone who's using an old version of the buildpack (or a branch) from accidentally using a non-GA version.

jkutner (approved):

data.json
danp: I can see that, but that's also true for something like "the default today is go1.12.1 but in this buildpack branch from 3 months ago it was go1.11.6," where go1.11.6 may have now-large security issues or something. I think letting people say `go1.13` and having that evolve to the GA and beyond versions is pretty reasonable.

jkutner: I think the difference is that an RC or beta is actually *before* 1.13, where as 1.12.1 is *after* 1.12 with an implication it might/should be backwards compatible. I don't think there's any guarantee that a beta or RC won't break before updating to GA. 

Ultimately, this kind of decision is up to the product owner though.

PR #366, opened 2019-09-25
Add go1.13.1, & go1.12.10 - default to go1.12.10

freeformz:


freeformz:
gb tests are failing with go1.13.1 so reverting back to go1.12 as the default. It's probably time to deprecate gb and skip these tests for go >= 1.13, but that can all be done in a follow up PR/release.

bin/test
```diff
@@ -40,6 +40,7 @@ else
   # When not using modules, $build is setup as the GOPATH by compile w/$GO_SETUP_GOPATH_IN_IMAGE
   export GOPATH="${build}"
   PATH="${GOPATH}/bin:${GOROOT}/bin:${PATH}"
+  export GO111MODULE=off
```
danp, line 43: Guessing this was part of changes to make go1.13 the default, still needed if this PR isn't doing that?

danp (approved):

bin/test
freeformz: it's probably still wise. I can remove it though if the consensus is that I should.

schneems (approved):

PR #371, opened 2019-10-11
Bump golangci-lint to 1.20.0

freeformz:
This fixes issues when using golangci-lint with projects that have moved to go1.13.


darron:
Looks great @freeformz - it works as designed now with our `1.13` apps.

Thanks for the quick response!

schneems (approved):

PR #392, opened 2020-02-26
add go1.14

danp:
https://golang.org/doc/go1.14

The file has been synced to S3, using #391.


danp:
Closing in favor of #393.

Issue #394, opened 2020-03-13
Ignore vendoring in Go 1.14


codedius:
Hi! I'm trying to build my app with Go 1.14, but faces the error:

> [module name] explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
run 'go mod vendor' to sync, or use -mod=mod or -mod=readonly to ignore the vendor directory

Is there any way to ignore Go vendoring in Heroku? I've tried use env GOFLAGS=-mod=mod but it didn't work.

LaysDragon:
I face the same problem while using gitlab auto devops 
need to find a way to get it use the go mod instead vendor

danp:
Hi there,

It sounds possible that `go mod vendor` hasn't been run locally with go1.14 yet. Could either of you please share the output of `go version` and `go env` from a project having this issue?

If I understand correctly, something like this might fix it:

1. Be sure `go version` reports go1.14 or greater
2. Within your project, run `go mod vendor`
3. Commit any changes, such as lines added to `vendor/modules.txt` prefixed with `#`
4. `git push heroku master` or similar

Thanks!

codedius:
Hi @danp,

Please, read the first message carefully. **WITH** vendoring everything works fine, but I'm interested in how to use Go 1.14 in Heroku **WITHOUT** vendoring. In simple words, I **don't want** to use `go mod vendor`, but I **have to**, because I get the error above otherwise. So, is there any way to ignore Go vendoring in Heroku?

danp:
Hi, sorry for misunderstanding. I could still use some info about the project you're working with to make a better recommendation. In addition to the info I asked for above, could you please provide the output of `ls -la vendor/modules.txt`?

Thanks!

codedius:
Ok, so:

1. `go version go1.14 darwin/amd64`
2. Done
3. Done
4. Done

Output:
`-rw-r--r--  1 lepeshev  staff  2560 Mar 12 18:52 vendor/modules.txt`

danp:
Thanks! How about `go env` output?

I'm also wondering if you have a line in `go.mod` that starts with `go` and what it says. If it doesn't say `go 1.14`, please update it and try the `go mod vendor` and following steps again.

In the meantime, I'll see if I can put together something that reproduces this problem so we can get it sorted.

Thanks again.

codedius:
Go env:

> GO111MODULE="on"
GOARCH="amd64"
GOBIN=""
[...]

PR #396, opened 2020-03-23
make `go list` use of -mod=vendor more dynamic

danp:
If a Go release of 1.14 or greater is used with a language version of
1.14 or greater, always passing -mod=vendor to `go list` breaks due to
new consistency checks in 1.14 (https://golang.org/doc/go1.14#vendor).

Maybe better would be to inspect the desired language version (using
`go mod edit -json | jq -r '.Go'` or so) but that could use some more
thinking.

To fix it for now, move up the adding of -mod=vendor to the build
flags and use those when calling `go list`. This should leave out
-mod=vendor if there is no vendor directory.

Updates #394 (will close once this is properly released)


kennyp (approved):
LGTM

jboursiquot-sfdc (approved):
LGTM

owenthereal (approved):
:+1:

schneems (approved):
LGTM

PR #399, opened 2020-03-27
compile: clear cache on stack change

danp:
It's helpful to clear the cache and start fresh when the [Heroku stack](https://devcenter.heroku.com/articles/stack) changes, such as going from heroku-16 to heroku-18. Especially when using external dependencies via cgo.

### DIFF
--- a/bin/compile
+++ b/bin/compile
@@ -13,6 +13,32 @@ source "${buildpack}/lib/common.sh"
 source "${buildpack}/lib/common_tools.sh"
 source_stdlib
 
+cacheStackFile="$cache/go-stack"
+
+cacheStack() {
+    echo "$STACK" > "$cacheStackFile"
+}
+
+resetCache() {
+    if [ -d "$cache/go-path" ]; then
+        find "$cache/go-path" ! -perm -u=w -print0 | xargs -r -0 chmod u+w 2>&1
+    fi
+    rm -rf "${cache:?}"/*
+    cacheStack
+}
+
+cachedStack=""
+if [ -r "$cacheStackFile" ]; then
+    cachedStack="$(< "$cacheStackFile")"
+fi
+if [ "$cachedStack" != "$STACK" ]; then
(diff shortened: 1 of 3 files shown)


danp:
This needs a tweak. A Go version change also wipes the cache and would clear the cached stack value.

bin/compile
```diff
@@ -261,7 +277,7 @@ ensureGo() {
         if [ -d "${cache}/go-path" ]; then
```
ghost: You are going to need to move these lines into `resetCache` as they guard against the read only nature of the go modules cache

bin/compile
```diff
@@ -13,6 +13,22 @@ source "${buildpack}/lib/common.sh"
 source "${buildpack}/lib/common_tools.sh"
 source_stdlib
 
+resetCache() {
+    rm -rf "${cache:?}"/*
+    echo "$STACK" > "$cache/stack"
```
ghost: What are the trickle-down effects if $STACK isn't set?

ghost: What are the effects of stacking this buildpack with other buildpacks that "reset the stack" ?

ghost (changes requested):

bin/compile
danp: Changed this up a little. Before, the stack value was only being cached if it changed. Now, it's always cached. If STACK is unset or empty, then an empty value is cached. I think that's okay, though, since changes will still be detected.

Re composing this buildpack with others, I also changed this one to caching the stack in `$cache/go-stack` instead of the more generic `$cache/stack`. I'm not sure what the expectations are around how buildpacks compose with each other, who owns the cache, etc. I think the changes I mentioned above will work well if another buildpack nukes the entire cache as this one does and has been.

It might be nice at some point to make an effort to better namespace this buildpack's use of the cache so it's easier to only clear its part, like putting everything under `$cache/go` or something.

PR #404, opened 2020-04-20
Add Heroku-20 to the Travis test matrix

edmorley:
To confirm that the buildpack works on the preview of the new stack.

The install step has also been updated to only pull the image being tested, rather than all of the images for each job in the matrix, saving 2 minutes per job.

I had to recompile [Yara 3.4.0](https://github.com/virustotal/yara/releases/tag/v3.4.0) in the `godep-cgo-vendored` fixture, since it was linked against `libssl1.0.0` which is not available for Ubuntu 20.04. 

Since there isn't a single libssl version that is supported across all of our stacks (`libssl1.1` is only available for Ubuntu 18.04+), I used `configure --without-crypto` to remove Yara's dependency on OpenSSL entirely.

The removed files were all redundant for the purposes of that test.

Refs [W-7472042](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07B0000007QrGTIA0/view).


danielleadams (approved):

.travis.yml
```diff
@@ -5,16 +5,15 @@ services:
 before_install:
 - echo -e "machine api.github.com\n  login $GITHUB_BASIC" > ~/.netrc
 install:
-- docker pull heroku/cedar:14
-- docker pull heroku/heroku:16-build
-- docker pull heroku/heroku:18-build
+- docker pull $IMAGE
```
danp, line 8: Nice.

Issue #377, opened 2019-10-24
support new go1.13 module proxy/sum/etc options


danp:
go1.13 made use of proxy.golang.org and others default, as mentioned [here](https://golang.org/cmd/go/#hdr-Module_downloading_and_verification).

We should give people a way to control values of GOPROXY, GOSUMDB, GOPRIVATE, and friends as part of builds. For example, apps using GitHub tokens to fetch from private repositories as part of builds may break or expose some information (repo name, etc).

ghost:
Internal Work Item: [W-6993106](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07B00000078TLRIA2/view)

jeffprestes:
I need this feature too. I can't deploy my new version as I can't make the buildpack download private libs from my bitbucket.org projects

jheth:
Yes please! We need this too.

PR #408, opened 2020-05-07
Non-default and private Go Modules proxy support

rykov:
We are now finalizing Go Modules support in Gemfury and would like to allow private dependencies to be used in Heroku projects from both the Gemfury add-on and our regular accounts.

To achieve this, a customer would set the variables GOPROXY, GOPRIVATE, and GONOPROXY for their application as described here: https://gemfury.com/guide/go/install-with-go/

This PR is to propagate these variables from the app config to the build environment directory, and have them be visible to the buildpack's `go build` command.

Related issue:
#377


rykov:
Bump. Any thoughts on this?

edmorley:
Requested in https://heroku.support/912962

rykov:
For reference, we're looking to enable this functionality for Heroku customer:
https://gemfury.com/guide/go-modules/install-with-go/

Edit: This article is Gemfury-specific, but applies to all non-default Go Modules proxies.

danp:
Thanks for this. We can get this merged if you'd be up for rebasing/merging in current main (especially for #427 to get CI working again).

danp:
Related tests would be welcome, as well, but we/I can come back around after and add those if you're not up for it.

rykov:
@danp Rebased to latest **main**, and I'll put together a test in a bit.

rykov:
@danp And now with the tests. Please let me know if you want access to the Gemfury account here, or if I should upload the source of that module somewhere.

Here's the tested dependency source:
https://gemfury.com/go-test-org/go:git.fury.io%2Fgo-test-org%2Fgotest

danp:
Thanks! If I get the right bits set up for the test in the buildpack's S3 bucket, would you be up for using that? That way the test wouldn't depend on another service and it would keep management of the data with us. It might be even better longer-term to have the tests use some kind of local fake module proxy to verify but we can work that out later.

danp:
Actually, going further: GOPROXY can be set to a filesystem path. Perhaps there's a way to include the module zip/etc in the fixture itself and verify that setting GOPROXY to the path fetches/uses the module as expected?

rykov:
@danp Although here at Gemfury, we're a little bit offended not to carry the same trust as GitHub and GitLab 🤣, we do understand your point of view - using a directory-repo is more stable. I'll update it in a bit.

rykov:
@danp Ok, the repo URL is now `file://...`.

danp:
Awesome, thanks! Looks like one more rebase/merge is in order to fix up CHANGELOG.md.

rykov:
> Awesome, thanks! Looks like one more rebase/merge is in order to fix up CHANGELOG.md.

Done

danp (approved):

edmorley (approved):

PR #409, opened 2020-05-11
[changelog skip] Fix PR edit check

schneems:
Previously when a PR title was edited, the "check for changelog" script would not fire. This means if someone edited it to add a "[changelog skip]" that the check would still show to be failing. This PR adds additional triggers for the check so that when the PR is edited, the check will re-run.


edmorley (approved):
Thank you for fixing this; it was driving me up the wall the other day! :-)

Issue #378, opened 2019-11-29
Use the go version in go.mod to set the minimal Go version to use


edmorley:
Currently this buildpack requires that the Go version be specified in a Heroku-specific format in `go.mod`, otherwise the build will output something like:

```
 !!    The go.mod file for this project does not specify a Go version
 !!    
 !!    Defaulting to go1.12.12
 !!    
 !!    For more details see: https://devcenter.heroku.com/articles/go-apps-with-modules#build-configuration
 !!    
```

At first glance it seems like the buildpack should just be using the `go 1.NN` style annotations now supported by Go in `go.mod`. 

However as discussed in #301, those annotations are meant to indicate the minimum language feature version required, and not necessarily the compiler version to use. (For example `go.mod` might say it requires `go 1.12`, which a go 1.13 compiler could support by running in "go 1.12 mode").

Whilst that may be true, I suspect in the common case, people just want their entire app to run using the same version in production, CI, ... - and will never need to use a mixture of language/compiler versions.

The current implementation (requiring Heroku-specific annotations) not only adds to the boilerplate required for projects using this buildpack, but can also mean projects unknowingly run a different version of Go in production.

For example in https://github.com/heroku/release-dyno-watcher/pull/191 that service was upgraded to Go 1.13, and whilst CI is now testing against Go 1.13, production is still running Go 1.12 per the buildpack default (which was not spotted since the build "The go.mod file for this project does not specify a Go version" warning is only a warning not fatal).

As such, would it be possible to either:
1. Have the standard `go.mod` version set the **minimal** version of the Go compiler to use (iff the user hasn't overridden with the `// +heroku goVersion` style syntax)
2. Have the standard `go.mod` version set the **specific major** version of the Go compiler to use (iff the user hasn't overridden with the `// +heroku goVersion` style syntax)

Whilst (1) might be closer to the intent of the upstream Go project (see https://github.com/heroku/heroku-buildpack-go/issues/301#issuecomment-475311555), I personally would prefer (2) since it offers the least surprise.

ghost:
I've re-read all of the posts involved with this directive that I can find and I think the only "safe" thing to do is, in the absence of the `// +heroku goVersion` directive, utilize an existing `go` directive as the **minimal major** Go version. This will allow users to specify, via the `// +heroku goVersion` directive, a newer Go version than the `go` directive indicates, allowing for a toolchain update w/o requiring a "language" version change, protecting against a future where that would be necessary.

[...]

PR #411, opened 2020-05-16
Use the go version in go.mod if no +heroku comment

abitrolly:
`go` records own version in `go.mod` as

    go 1.13

This fix should let the buildpack to detect it if nothing else is set.
https://github.com/heroku/heroku-buildpack-go/issues/378#issuecomment-629665034

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,6 +52,7 @@
 ## v155 (2021-09-13)
 * Add go1.16.8, use for go1.16
 * Add go1.17.1, use for go1.17
+* Use the go version in go.mod if no +heroku comment is found (#378)
 
 ## v154 (2021-08-18)
 * Add go1.17
--- a/data.json
+++ b/data.json
@@ -118,6 +118,7 @@
       "go1.10.8.linux-amd64.tar.gz",
       "go1.11.13.linux-amd64.tar.gz",
       "go1.12.17.linux-amd64.tar.gz",
+      "go1.13.15.linux-amd64.tar.gz",
       "go1.14.2.linux-amd64.tar.gz",
       "go1.15.15.linux-amd64.tar.gz",
       "go1.16rc1.linux-amd64.tar.gz",
--- a/lib/common.sh
+++ b/lib/common.sh
@@ -330,6 +330,7 @@ determineTool() {
         info "Detected go modules via go.mod"
         step ""
         ver=${GOVERSION:-$(awk '{ if ($1 == "//" && $2 == "+heroku" && $3 == "goVersion" ) { print $4; exit } }' ${goMOD})}
+        ver=${ver:-$(awk '{ if ($1 == "go" ) { print "go" $2; exit } }' ${goMOD})} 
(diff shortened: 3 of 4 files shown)


abitrolly:
No idea what is going on and why tests fail.

UebelAndre:
Any updates here?

frederikhors:
This is a good idea I found searching issues!

Please use this, team!

tt (dismissed):
The tests seem to fail because parsing the Go version from go.mod causes it to use Go 1.13.14 rather than the default 1.12.17 for the two fixtures and that version is not available for tests.

If you add "go1.13.14.linux-amd64.tar.gz" in data.json under /test/assets, you should be able to remove the two last commits, 489a544a8830c2a71e49de78dd7776f283aede9d and 2328a677c72404747142324cd70333f267ac09f5, and see the tests pass.

I'm happy to make the changes for you. Let me know if you prefer that.

abitrolly:
@tt please do. I am overwhelmed with failing migration from Rails 4 to Rails 5 on my Heroku project, and won't be able to get back to this PR soon.

bchatelard:
Any updates ?

abitrolly:
@tt rebased and removed extra commits. Now only one tests is failing.

```
testModPrivateProxy
* fixture: mod-private-proxy/app
* detect
* compile
ASSERT:Expected </tmp/shunit.DrxOhj/tmp/output.PwTF/stdout> to contain <Installing go1.12.17
Fetching go1.12.17.linux-amd64.tar.gz... done>
testModDeps
* fixture: mod-deps
* detect
* compile
```

This one is tricky. I could downgrade the Go 1.15 from the fixture's `go.mod` to Go 1.12, but I am not sure Go 1.12 has features that this part is testing.

abitrolly:
@edmorley @joshwlewis can you take a look?

abitrolly:
Rebased to see if the problem with tests is gone.

@tt looks like it is blocked on your feedback or approval to continue.

kennyp:
So the semantic meaning of the [go directive](https://go.dev/ref/mod#go-mod-file-go) is not actually the same as the `goVersion` so the behavior in this case would be a bit different that what the go docs imply. What's the reason for not wanting to include the directive and be explicit?

abitrolly:
@kennyp is it the question for me?

joshwlewis:
> So the semantic meaning of the [go directive](https://go.dev/ref/mod#go-mod-file-go) is not actually the same as the goVersion so the behavior in this case would be a bit different that what the go docs imply. 

Yes, the semantic meaning is slightly different. But I think this is a better default than the buildpack global default, which is currently `go1.12.17`. `go 1.12` seems pretty dated in any app with a `go.mod`. The global default will be a little easier to change, if we know it will affect a smaller number of apps (those without any more explicit version constraints).

> What's the reason for not wanting to include the directive and be explicit?

I think it's just ergonomics and convenience. We can pick a better default for folks, and improve the first deploy experience. Being explicit is still possible here.

joshwlewis (approved):

abitrolly:
Thanks for merging.

PR #420, opened 2020-07-25
Git submodules support

pascencio:
Hi,

I'm trying to build my application and deploy it in heroku. I want divide my application in back y front using git submodules. The problem begins when golang buildpack try to check if the project use go.mod or other tool. In my project use standard go modules and the `common.sh` look for `go.mod` file in the root of project (or `${build}` path).

In my branch change the way to get `go.mod` file path and the installation process in `compile` shellscript.

I hope that my contribution be helpful.

Thanks in the advance!

### DIFF
--- a/bin/compile
+++ b/bin/compile
@@ -112,7 +112,7 @@ expandVer() {
 reportVer() {
     local ver="${1}"
     mcount "version.go.$ver"
-
+    
     case $ver in
         devel*)
             warn ""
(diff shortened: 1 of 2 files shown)


edmorley (commented):

joshwlewis:
It's great that this works for you. I think this would need more work to enable it for a wider range of apps. It also needs tests, and I would prefer not to see the whitespace changes.

PR #440, opened 2020-12-31
support go1.16beta1

axetroy:


axetroy:
is there something wrong with changelog?

files.json
```diff
@@ -467,6 +467,14 @@
     "SHA": "2d75848ac606061efe52a8068d0e647b35ce487a15bb52272c427df485193602",
     "URL": "https://storage.googleapis.com/golang/go1.15.linux-amd64.tar.gz"
   },
+  "go1.16.linux-amd64.tar.gz": {
```
danp, line 470: Please remove this, files should only be listed here as they appear at https://golang.org/dl/ or similar. Using 1.16beta1 when `go1.16` is requested is controlled by data.json.

data.json
```diff
@@ -2,6 +2,8 @@
   "Go": {
     "DefaultVersion": "go1.12.17",
     "VersionExpansion": {
+      "go1.16.0": "go1.16",
```
danp, line 5:
```suggestion
      "go1.16.0": "go1.16beta1",
```

When go1.16 comes out, this will be updated to go1.16 and stay there. The `go1.16` key on the next line will track the latest go1.16.x version.

danp (commented):
Thanks for this!

Adding an entry to CHANGELOG.md under the Unreleased section will fix the check, something like:

```
* Add go1.16beta1, use for go1.16
```

tt:
I'm closing this as #449 was merged a while ago.

Issue #436, opened 2020-12-02
Allow specifying the golangci-lint version


djgilcrease:
This is currently hardcoded to 1.20.0 and I would like to use a newer version as v1.33.0 is currently the latest. Maybe make the install command something like;

`curl -sfL https://install.goreleaser.com/github.com/golangci/golangci-lint.sh | sh -s -- -b ${build}/.heroku/golangci/bin ${GOLANCI_VERSION:v1.20.0}`

That way you do not need to create a new release every time a new version of `golangci-lint` is released.

PR #459, opened 2021-03-16
bump gitlabci-lint version to 1.38

GRbit:
When users use newer golngci config on local machines, running the old version on Heroku may fail with this kind of errors
```
-----> Running: golangci-lint -v --build-tags heroku run
level=info msg="[config_reader] Config search paths: [./ /app /]"
level=info msg="[config_reader] Used config file .golangci.yml"
level=error msg="Running error: no such linter \"gomnd\""
level=info msg="Memory: 2 samples, avg is 68.4MB, max is 68.4MB"
level=info msg="Execution took 15.484269ms"
```

So it's important to keep latest version of golangci-lint, or even [allow specifying the golangci-lint version](https://github.com/heroku/heroku-buildpack-go/issues/436) since it can be downloaded directly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Go Buildpack Changelog
 
 ## Unreleased
+* Bump golangci-lint to v1.38.0
 
 ## v153 (2021-03-11)
 * Add go1.16.1, use for go1.16
@@ -652,3 +653,4 @@
 ## v12 (2015-06-29)
 
 * [GOPATH naming changed & update godep](https://github.com/heroku/heroku-buildpack-go/pull/82)
+
--- a/bin/test-compile
+++ b/bin/test-compile
@@ -24,5 +24,5 @@ if [ -f "${build}/.golangci.yml" -o -f "${build}/.golangci.toml" -o -f "${build}
     step "/.golangci.{yml,toml,json} detected"
     tmp="$(mktemp -d)"
     mkdir -p "${build}/.heroku/golangci/bin"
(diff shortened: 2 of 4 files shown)


GRbit:
I've make my PR basing on https://github.com/heroku/heroku-buildpack-go/pull/371, it seems pretty straightforward and I don't know why tests are failing.

Tried to run my project with my buildpack, it stops on this error
```
-----> Copying go tool chain to $GOROOT=$HOME/.heroku/go... done
-----> /.golangci.{yml,toml,json} detected
-----> Fetching golangci-lint-1.38.0-linux-amd64.tar.gz... 
gzip: stdin: not in gzip format
tar: Child returned status 1
tar: Error is not recoverable: exiting now
```

Tried to download file manually and unpack it with specified tar command - everything working just fine. Have no idea what can possibly go wrong on Heroku side.

Can someone clarify this situation?

kennyp:
@GRbit there's actually a caching step we do on the back-end so we would need to add the new version to the cache.

GRbit:
> @GRbit there's actually a caching step we do on the back-end so we would need to add the new version to the cache.

Thank you for the answer. My PR is outdated, now it's time to update golangci-lint to the latest version. Can I help somehow? As I understood, I have no options to update your caching, so I can only update golangci version. Should I?

kennyp:
@GRbit I just cached `1.45.0`. If you want to update your PR to include that instead of `1.38`. Also, add `1.16` and `1.20` back in, since we don't delete old versions in case we need to rollback to a previous buildpack version.

edmorley (commented):

joshwlewis:
Closing this due to age.

PR #460, opened 2021-03-17
Test version expansion

tt:
This makes sure to test that version expansion for the supported versions (the two latest) work which will also ensure that the binaries are available in the appropriate S3 bucket.


edmorley (approved):
Thank you for this! :-)

If you didn't want the changelog addition (since this is test-only), another option would be to add `[changelog skip]` to the PR title (doesn't need to be on the commit itself), which would cause the changelog check to be skipped. (Some of the other repos support labels and/or PR description for this, to avoid cluttering the PR title, but this repo's GH workflow config doesn't unfortunately.)

edmorley:
Travis is failing

data.json
```diff
@@ -115,7 +115,9 @@
       "go1.11.13.linux-amd64.tar.gz",
       "go1.12.17.linux-amd64.tar.gz",
       "go1.14.2.linux-amd64.tar.gz",
+      "go1.15.9.linux-amd64.tar.gz",
```
tt, line 118: This is an unfortunate side effect of requiring test assets to be declared. Testing version expansion this way requires the file to be changed whenever new Go versions are added and will require downloading a couple hundred megabytes more before tests can run.

edmorley: Yeah not ideal, though think still worth adding these tests for now and fixing the way tests work later. (Ideally this repo would use Hatchet tests for much of this anyway.)

edmorley:
@tt Could you rebase this, now that #461 has merged?

edmorley:
Thank you :-)

PR #463, opened 2021-05-05
Set xtrace if $BUILDPACK_XTRACE set

josegonzalez:
For debugging or for folks who are curious about how the buildpack works.

Similar implementation here: https://github.com/heroku/heroku-buildpack-python/pull/226

### DIFF
--- a/bin/compile
+++ b/bin/compile
@@ -3,6 +3,8 @@
 
 set -eo pipefail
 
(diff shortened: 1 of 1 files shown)


kennyp:
@heroku/languages this is inline with how we're doing things in the Python buildpack. Any reason not to merge?

joshwlewis (approved):

PR #464, opened 2021-05-05
don't hide error messages from go list

fabjan:
## Why?

I had a repo in a bad state (I missed comitting an updated go.sum) which failed to build but figuring out why took some time because the error messages were hidden.

## Result

Before this change:
```
-----> Determining packages to install
remote: 2021/05/05 21:32:22 exit status 1
```

After this change:
```
-----> Determining packages to install
remote: go: github.com/jackc/pgx/v4@v4.11.0: missing go.sum entry; to add it:
remote:         go mod download github.com/jackc/pgx/v4
remote: 2021/05/05 21:41:10 exit status 1
```

That output makes identifying my error much quicker.

When go.sum is fixed there is no extra error output from go list due to this change.

## Comments

The /dev/null redirection was added with the commit "[make go list use of -mod=vendor more dynamic]", which does not mention anything about the error output, so if it was not strictly necessary maybe that little redirection change can be reverted. I see a few parts of the script redirecting to stdout as well but go list did just not do any redirection prior to that commit.

Also, `jq` is required for running `make test`. I can add that info to the README in this PR if you want.

[make go list use of -mod=vendor more dynamic]: https://github.com/heroku/heroku-buildpack-go/commit/fb22bbe8fb3d9605507a8d889648ee98c40d57b6#diff-a64a7d9b82578ca3053416d56631a61e0ceabed87619415663030e3842429e05R427


fabjan:
Running tests I see that non-error cases also seem to write to stderr making tests fail ... I should have let them run to completion before creating the PR.

joshwlewis (approved):
The big change to me here is the additional `downloading git.fury.io/go-test-org/gotest v0.1.0` type lines in stderr. My guess is that these are the reason for the redirection in the first place, as they'll get noisy with bigger dependency chains.

However, I think most folks are used to these messages from local development, the messages can be helpful to understand what the buildpack is doing, and the error message that was previously masked should be in the build logs.

Would you mind adding an entry to the `Unreleased` section of the `CHANGELOG.md` describing this change?

fabjan:
@joshwlewis my bad, I can resolve the conflict. Is a rebase in my fork the preferred way?

Also, I added a note about jq being required to run the tests (I think because the `test` target depends on `test-assets` which runs `sbin/fetch-test-assets` which calls out to `jq`). I can drop that change from this PR if you prefer.

fabjan:
conflicts resolved

joshwlewis (approved):

PR #478, opened 2021-11-15
go1.16.10 and go1.17.3

thepwagner:
Makes the latest releases available, as announced https://groups.google.com/g/golang-announce/c/0fM21h43arc .

Checksums can be validated at https://golang.org/dl/?mode=json


thepwagner:
I've double checked this diff, and I'm not sure how the test failures are related.

joshwlewis (approved):
Thanks! The CI failures relate to outside collaborators not having access to private resources. I just had to retrigger CI myself.

PR #484, opened 2022-03-18
add workspace support

wexder:
This adds support for new go workspace file. It should behave the same as when using go modules.

### DIFF
--- a/bin/compile
+++ b/bin/compile
@@ -42,6 +42,11 @@ handleDefaultPkgSpec() {
                 warn "// +heroku install ./cmd/..."
                 warn ""
             ;;
+            goworkspaces)
+                warn "To install a different package spec add a comment in the following form to your \`go.work\` file:"
+                warn "// +heroku install ./cmd/..."
+                warn ""
+            ;;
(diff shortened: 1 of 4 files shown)


edmorley (commented):

bin/compile
```diff
@@ -533,6 +538,77 @@ case "${TOOL}" in
             bin/go-post-compile
         fi
     ;;
+    goworkspaces)
```
joshwlewis, line 541: This looks like an exact copy of the gomodules section? I can't easily tell what's different. Is there a way to reduce the duplication here? Maybe goworkspaces and gomodules could share the same branch of the case statement?

joshwlewis (commented):
If you are still interested in pursuing this PR, it'll need some tests of the new functionality.

wexder:
> If you are still interested in pursuing this PR, it'll need some tests of the new functionality.

I'm no longer working with heroku so I'm not interested in finishing this. Anyone is free to pick up the work where I left it, but It looks like noone needed this is last 2 years...

PR #487, opened 2022-05-19
Update Go version inventory

joshwlewis:
New Versions:

- `1.18.1`
- `1.18.2`
- `1.17.9`
- `1.17.10`

Default changes:

- `1.18: 1.18 => 1.18.2`
- `1.17: 1.17.8 => 1.17.10`

PR #488, opened 2022-06-01
Fix CHANGELOG entry for go.mod parsing in #411 and add missing v162 release

abitrolly:


edmorley (approved):
Thank you for spotting this!

I also noticed that the v162 release has already occurred:

```
$ h buildpacks:versions heroku/go | head -n4
Version  Released At               Status
───────  ────────────────────────  ─────────
162      2022-05-23T19:42:06.208Z  published
161      2022-03-15T22:11:28.510Z  published
```

The changelog was just not updated after #487 was published.

I've added commit 5b5f865f927e37ca23631a97095a0ae91e4cd760 to this PR which adds that undocumented release, and moves the changelog entry for #411 to the unreleased section.

PR #489, opened 2022-06-07
Add support for 1.18.3 and make it default for 1.18

nestorsalceda:
New Versions:

* `1.18.3`

Default changes:

* `1.18: 1.18.2 => 1.18.3`

PR #490, opened 2022-06-09
Release v163

joshwlewis:
Adds the changelog entry for v163 that we can ship today.


edmorley (approved):

edmorley:
@joshwlewis Do we need a https://devcenter.heroku.com/changelog entry for #489 + #411? :-)

PR #521, opened 2023-08-11
add go1.21

viczaca:
This adds go1.21 support to the buildpack

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 # Go Buildpack Changelog
 
 ## Unreleased
+* Add go1.21.0
+* go1.21 defaults to go1.21.0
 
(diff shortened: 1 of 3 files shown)


joshwlewis (changes requested):
Hey, thanks. This looks great, but we need a test and fixture that exercises go1.21. Really, copying the 1.20 test/fixture forward for 1.21 would be fine: https://github.com/heroku/heroku-buildpack-go/blob/7d4f622f794f31a2941f8ba854e149b5cb21aa13/test/run.sh#L357.

joshwlewis:
Hey @viczaca - thanks for the contribution. I went ahead and brought in go1.21 in with #522, since we're a few days past the release by now.

PR #522, opened 2023-08-14
Add go1.21, use for go1.21.0

joshwlewis:
This adds go1.21.0, and uses go1.21.0 for go1.21. It also adds a simple test for the go1.21 line.

Supersedes #521.


edmorley (approved):
The change looks good, but CI is failing with `curl: (6) Could not resolve host: null` even on retry - an unrelated issue, but strange nonetheless?

edmorley:
Oh, it seemed unrelated because there were a dozen error lines (so it seemed multiple assets were failing the same way), but I guess the additional lines were just retries, so they were all from the newly added assets - whose URLs have just been fixed in the followup commits?

joshwlewis:
> The change looks good, but CI is failing with curl: (6) Could not resolve host: null even on retry - an unrelated issue, but strange nonetheless?

I think this has to do with an upstream artifact naming change. Traditionally, the first release doesn't have a `.0` suffix for go releases. For example, the first `go1.20` was `go1.20` (no `.0`), and `go1.20.0` doesn't exist in the upstream distribution repository. For some reason, there is a suffix on this release - `go1.21` doesn't exist upstream, but `go1.21.0` does.

joshwlewis:
It looks like the `.0` is intentional: https://github.com/golang/go/issues/57631

PR #532, opened 2024-01-02
Prepare release v183

joshwlewis:
This will mark the release of go 1.20.12 and 1.21.5 support with v183.


edmorley (approved):
(Approving since whilst the prod change moratorium doesn't end until tomorrow, we have a moratorium exception for lang updates)

PR #536, opened 2024-01-26
Use the new "Prepare release" GitHub Actions workflow

edmorley:
We now have a new reusable GitHub Actions workflow for preparing a new classic buildpack release:
https://github.com/heroku/languages-github-actions/blob/main/.github/workflows/_classic-buildpack-prepare-release.yml

This workflow works similarly to the CNB prepare release workflow, except it does not need any version bump input, since we only ever increment classic buildpack versions by one.

GUS-W-14901966.

### DIFF
--- a/.github/workflows/prepare-release.yml
+++ b/.github/workflows/prepare-release.yml
@@ -1,0 +1,12 @@
+name: Prepare Release
+
+on:
+  workflow_dispatch:
+
+# Disable all GITHUB_TOKEN permissions, since the GitHub App token is used instead.
+permissions: {}
+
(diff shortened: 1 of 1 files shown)


joshwlewis (approved):
Looks great. I'll ship it next week!

edmorley:
The change moratorium ended today at 13:00 UTC, so I've merged this now, so the new workflow is ready for whenever you'd like to perform the next release (no rush from my side, that can be next week if preferred):
https://github.com/heroku/heroku-buildpack-go/actions/workflows/prepare-release.yml

PR #558, opened 2024-06-04
Add go1.22.4 and go1.22.11

joshwlewis:
This adds go1.22.4 and go1.22.11, and makes them the default for their respective lines. It also tags the v192 release which will contain only these new versions.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## [Unreleased]
 
+## [v192] - 2024-06-04
+
 * Add go1.22.4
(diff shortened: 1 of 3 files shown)


runesoerensen (approved):

HCoban:
Hi @joshwlewis  

I'm facing build failures due to go1.22.4 not being able to be pulled. I suspect it might be related to this change. I created [an issue](https://github.com/heroku/heroku-buildpack-go/issues/559). 

I would appreciate your help. Thank you very much in advance!

PR #561, opened 2024-06-25
Source GOFLAGS from environment

### LINKED ISSUE #400: Further specifying build flags
Hello,

I can't find any way of passing custom build flags to the Go compiler.
I tried using the `GOFLAGS` environment with `heroku config:set GOFLAGS="my extra flags"` but I doesn't work.

Ideally, I need a solution to pre-install a tool that I pass to the compiler through its flags:
1. `go install my-tool`
2. `go build <my options with my-tool> app`

Is it possible? I will otherwise have to move to a Dockerfile solution for now.
Or maybe should I create a separate buildpack for that :thinking: ?

For some context: I am the tool provider, willing to document the Heroku use-case ;-)
Thanks!

joshwlewis:
This allows users to customize their builds a bit more, and do things like `heroku config:set GOFLAGS="-pgo=/tmp/profile.pgo"` so that PGO may be used.

Fixes #400.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## [Unreleased]
 
+* Source GOFLAGS from the environment
+
 ## [v192] - 2024-06-04
 
 * Add go1.22.4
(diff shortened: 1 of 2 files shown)


ypaq (approved):
Thank you!

edmorley (approved):

PR #578, opened 2024-11-12
fix: remove '-tags heroku' passed to CLI to honor '-tags' in GOFLAGS

GuillaumeDesforges:
The CLI flag `-tags heroku` forcefully set in the buildpack overrides the tags set in the GOFLAGS environment variable, as per specified in:

https://github.com/golang/go/blob/c96939fbed3d60159dc81ee9ad591de8cfd41168/src/cmd/go/internal/help/helpdoc.go#L525-L531

> Flags listed on the command line are applied after this list and therefore override it.

However, it makes it impossible to set other tags such as `viper_bind_struct` (see https://github.com/spf13/viper/pull/1720) using the standard `GOFLAGS="-tags=viper_bind_struct"`.

### DIFF
--- a/bin/compile
+++ b/bin/compile
@@ -393,8 +393,6 @@ loadEnvDir "${env_dir}"
 setGitCredHelper "${env_dir}"
 trap clearGitCredHelper INT TERM EXIT
 
-FLAGS=(-tags heroku)
-
 determineTool
 
 ver=$(expandVer $ver)


joshwlewis (changes requested):
There's a long history of apps on Heroku that depend on the `-tags heroku` for conditional compilation. Removing this would break a bunch of apps, so I can't accept this as-is. Maybe the buildpack could set a default value of `GOFLAGS=-tags heroku` when GOFLAGS was empty or unset instead?

GuillaumeDesforges:
Makes sense.

I'm not using this anymore so I will leave it at that.

PR #587, opened 2025-02-12
Add go1.24.0

joshwlewis:
Adds go1.24.0 and makes it the default for go1.24.

### DIFF
--- a/data.json
+++ b/data.json
@@ -2,6 +2,7 @@
   "Go": {
     "DefaultVersion": "go1.20.14",
     "VersionExpansion": {
+      "go1.24": "go1.24.0",
       "go1.23": "go1.23.6",
       "go1.22": "go1.22.12",
       "go1.21": "go1.21.13",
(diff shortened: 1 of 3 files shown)


data.json
```diff
@@ -134,6 +135,7 @@
       "go1.21.13.linux-amd64.tar.gz",
       "go1.22.12.linux-amd64.tar.gz",
       "go1.23.6.linux-amd64.tar.gz",
+      "go1.24.6.linux-amd64.tar.gz",
```
edmorley: I believe this is the cause of the strange `Fetching null` error in the CI run? 

```suggestion
      "go1.24.0.linux-amd64.tar.gz",
```

I wonder perhaps whether adding `--exit-status` to the jq usage here might bubble that error up sooner and make it easier to understand from the CI logs?
https://github.com/heroku/heroku-buildpack-go/blob/61b3a567c7a1d61617f50160f3a051616ddecfc7/sbin/fetch-test-assets#L41

xref https://jqlang.org/manual/#invoking-jq

edmorley (approved):

data.json
```diff
@@ -2,6 +2,7 @@
   "Go": {
     "DefaultVersion": "go1.20.14",
```
edmorley, line 3: (not a request for this PR, since unrelated)

The buildpack default Go is v1.20, which if I'm reading the version policy is unsupported and thus potentially insecure?
https://go.dev/doc/devel/release#policy

joshwlewis: Yes, that's true. Though prior to 2024, the default was almost never changed, and as a result, almost nobody used our default. 

We don't have a policy for updating the default, unlike most of our other supported languages. I think favoring the oldest supported makes sense though.

PR #609, opened 2025-10-07
Add go1.24.8 and go1.25.2 

runesoerensen:
GUS-W-19841008


Malax (approved):
I manually verified the checksums and they check out. :)

PR #623, opened 2026-01-17
Implement Go version update script and GHA workflow

runesoerensen:
This PR introduces automation to fetch and apply new Go versions from `go.dev` via a script and scheduled GitHub Actions workflow.

GUS-W-20996304


sbin/update-go-versions
```diff
@@ -0,0 +1,76 @@
...
+    echo ""
+    echo "Examples:"
+    echo "  $0 1.24.12,1.25.6    # Add specific versions"
+    echo "  $0                   # Auto-detect and add latest stable versions"
+    echo ""
+    echo "No versions specified, fetching latest stable versions from go.dev..."
+
+    # Fetch latest stable versions from go.dev
+    latest_versions=$(curl -s 'https://go.dev/dl/?mode=json&include=stable' | \
+        jq -r '.[].version | sub("^go"; "")')
```
schneems: Catch network errors on failure (add `-f`): 

```suggestion
    latest_versions=$(curl -sf 'https://go.dev/dl/?mode=json&include=stable' | \
        jq -r '.[].version | sub("^go"; "")')
```

sbin/update-go-versions
```diff
@@ -0,0 +1,76 @@
+#!/bin/bash
+
+set -e
```
schneems: More strict check

```suggestion
set -euo pipefail
```

sbin/update-go-versions
```diff
@@ -0,0 +1,76 @@
...
+
+cd "$(dirname "${BASH_SOURCE[0]}")/.."
+
+if [ -z "$1" ]; then
```
schneems: Required if we make the`set` check stricter

```suggestion
if [ -z "${1:-}" ]; then
```

schneems (approved):
Suggested optional: Add a shellcheck to CI and add this

sbin/update-go-versions
runesoerensen: Probably what you meant, but just in case: `curl` already returns a non-zero exit code for network errors, `--fail` makes it do it for HTTP errors as well :).

And thanks!

sbin/update-go-versions
```diff
@@ -0,0 +1,76 @@
...
+    echo "Usage: $0 [comma-separated-versions]"
+    echo ""
+    echo "Examples:"
+    echo "  $0 1.24.12,1.25.6    # Add specific versions"
+    echo "  $0                   # Auto-detect and add latest stable versions"
+    echo ""
+    echo "No versions specified, fetching latest stable versions from go.dev..."
+
+    # Fetch latest stable versions from go.dev
+    latest_versions=$(curl -sf 'https://go.dev/dl/?mode=json&include=stable' | \
```
edmorley, line 17: curl with `-s` but not `-S` hides all errors, which makes debugging harder.

Also, IMO we should always be using the other best practices when calling curl (eg timeouts, retries). Obviously this use-case is non-customer critical - though failures will still delay a Go update PR being opened by 24 hours, so I think it's still worth making this script robust :-)

runesoerensen: Good point -- pushed a PR to make it more robust https://github.com/heroku/heroku-buildpack-go/pull/628

PR #624, opened 2026-01-22
Verify test assets against S3 mirror

runesoerensen:
This PR updates the `sbin/fetch-test-assets` script to verify that cached or downloaded test assets exist in our S3 mirror, and that the local file size matches the `Content-Length` of the S3 object (without downloading the full file).

Previously, the script only verified SHAs against `files.json`, allowing tests to pass even if assets (like Go runtimes) were missing from the S3 mirror. This adds a basic sanity check to prevent tests from passing locally or in CI, when the buildpack would fail in production due to unsynced or corrupted artifacts.

### Example Output

Running with a newly added test asset (that exists in `files.json` and is available upstream, but hasn't been uploaded to our S3 mirror):

```console
$ make test-assets
Setting up test assets
Fetching https://dl.google.com/go/go1.26rc2.linux-amd64.tar.gz
Error: File missing from mirror: https://heroku-golang-prod.s3.us-east-1.amazonaws.com/go1.26rc2.linux-amd64.tar.gz
make: *** [test-assets] Error 1
```

Running with a `test/assets/go1.14.2.linux-amd64.tar.gz` file that has a different size than the one in our S3 mirror:

```console
$ make test-assets
Setting up test assets
Error: Mirror size mismatch for go1.14.2.linux-amd64.tar.gz (Local: 120173208, Mirror: 123658438)
make: *** [test-assets] Error 1
```

GUS-W-14293074


schneems (approved):
Seems good, thanks. We're missing pipefail from this script (pre-existing issue). It would be good to add.

PR #628, opened 2026-01-27
Improve robustness of Go version fetching

runesoerensen:
Add timeouts, retries, and connection refusal handling to the `curl` command fetching new versions from `go.dev`.

See related review for context https://github.com/heroku/heroku-buildpack-go/pull/623#discussion_r2731214206.

GUS-W-21056910


edmorley (approved):
Thank you :-)

PR #643, opened 2026-02-25
Switch to AWS' dual-stack S3 URLs

runesoerensen:
The default AWS S3 URLs only support IPv4, whereas the dual-stack URLs support both IPv4 and IPv6:
https://docs.aws.amazon.com/AmazonS3/latest/API/ipv6-access.html
https://docs.aws.amazon.com/AmazonS3/latest/API/dual-stack-endpoints.html

By switching to the dual-stack URLs, it means systems that support IPv6 will now use it instead of IPv4, which is helpful in IPv6 environments where IPv4 traffic has to fall back to being routed via NAT gateways (which has performance and ingress cost implications).

GUS-W-21336853

### DIFF
--- a/lib/common.sh
+++ b/lib/common.sh
@@ -19,7 +19,7 @@ NC='\033[0m' # No Color
 CURL="curl --silent --show-error --location --fail --retry 15 --retry-delay 2 --retry-connrefused --connect-timeout 5" # retry for up to 30 seconds
 
 if [ -z "${GO_BUCKET_URL}" ]; then
-    BucketURL="https://heroku-golang-prod.s3.us-east-1.amazonaws.com"
+    BucketURL="https://heroku-golang-prod.s3.dualstack.us-east-1.amazonaws.com"
 else
     BucketURL="${GO_BUCKET_URL}"
 fi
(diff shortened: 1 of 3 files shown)


lib/common.sh
```diff
@@ -19,7 +19,7 @@ NC='\033[0m' # No Color
 CURL="curl --silent --show-error --location --fail --retry 15 --retry-delay 2 --retry-connrefused --connect-timeout 5" # retry for up to 30 seconds
 
 if [ -z "${GO_BUCKET_URL}" ]; then
-    BucketURL="https://heroku-golang-prod.s3.us-east-1.amazonaws.com"
+    BucketURL="https://heroku-golang-prod.s3.dualstack.us-east-1.amazonaws.com"
```
edmorley, line 22: While it's only used in testing (so not used in production environments which might be behind a NAT gateway), I think it might be worth also changing the URL here:
https://github.com/heroku/heroku-buildpack-go/blob/76eca13da7a48172c5f4fcf5a96fa7d7ce9d4f97/sbin/fetch-test-assets#L37

...if only so (a) there's parity from a dev-prod point of view, (b) others in the future aren't wondering why two different URL styles are used and thinking there's more meaning behind the divergence than there is.

edmorley (approved):

PR #636, opened 2026-02-13
Add initial `bin/report` implementation

runesoerensen:
This PR adds `bin/report` to output structured build metadata for observability, tracking Go version resolution and dependency manager detection. Adapted from the [Python buildpack's implementation](https://github.com/heroku/heroku-buildpack-python/blob/8eaa83cf0f9e196abd13e895bc667274183909a6/bin/report).

_**Note:** Build data tracking is implemented directly within `determineTool()` to ensure metadata is captured even when the function exits early (e.g., validation errors). This lets `bin/report` provide observability data for failed builds, and shouldn't have any impact on the current logic/build output. While the existing logic could certainly be refactored and improved further, this approach should be sufficient for gathering usage data to inform the planned removal of deprecated tools (dep, godep, govendor, glide, gb) in favor of Go modules._

Depends on https://github.com/heroku/heroku-buildpack-go/pull/635

GUS-W-21257738

### New Files

**`lib/build_data.sh`** - Adapted from [Python buildpack's `lib/build_data.sh`](https://github.com/heroku/heroku-buildpack-python/blob/8eaa83cf0f9e196abd13e895bc667274183909a6/lib/build_data.sh)
- Includes functions for storing strings, raw JSON, and durations
- Stores data in `${CACHE_DIR}/build-data/go.json`
- Preserves previous build data as `go-prev.json` (which we may want to use in the future for cache invalidation etc)

**`bin/report`** - Adapted from [Python buildpack's `bin/report`](https://github.com/heroku/heroku-buildpack-python/blob/8eaa83cf0f9e196abd13e895bc667274183909a6/bin/report)
- Outputs sorted JSON of all tracked build metadata
- Called by the build system after `bin/compile` (including on failures)

### Modified Files

**`bin/compile`**
- Initialize build data store at start
- Track expanded Go version and pinned status after `determineTool()`:
  - `go_version`: Resolved version after expansion (e.g., `go1.21.6`)
  - `go_version_pinned`: Whether version was fully specified (e.g., `go1.21.13` is pinned, `go1.21` is not)
- Removed duplicate `setGoVersionFromEnvironment()` function definition (now uses the one from `lib/common.sh`)

**`lib/common.sh`**
- Refactored `determineTool()` to track metadata as data is discovered:
  - `go_tool`: Package manager detected (`gomodules`, `dep`, `godep`, `govendor`, `glide`, `gb`)
  - `go_version_requested`: User-specified version before expansion (e.g., `go1.21`)
[...]

PR #644, opened 2026-02-26
Remove support for legacy Go tools and versions

runesoerensen:
All Go dependency managers except Go modules have been [deprecated since March 1, 2025](https://devcenter.heroku.com/changelog-items/2983). This PR removes support for dep, godep, govendor, glide, and gb from the classic buildpack, making `go.mod` the only supported dependency management solution. It also removes < go1.11 Go artifacts (unused since modules require >= go1.11), simplifies SHA validation, and updates documentation.

Apps using legacy dependency managers (dep, godep, govendor, glide, gb) will still pass `bin/detect` - following the [Python buildpack's approach](https://github.com/heroku/heroku-buildpack-go/pull/644#discussion_r2863488784), detection is lenient so that `bin/compile` can provide a specific "support for X has been removed" migration error rather than the generic "can't detect language" message. The detected legacy tool is also recorded in build data (`go_tool`) for observability.

Apps that have a `go.mod` but specify a Go version < go1.11 (e.g. via `// +heroku goVersion go1.10` or `GOVERSION` env var) will continue to fail at compile with a "does not support Go modules" error — `bin/report` still runs and outputs `go_version_requested`, which (along with the build status) can be used to detect this failure mode.

These changes should have minimal impact on apps currently using this buildpack. Since https://github.com/heroku/heroku-buildpack-go/pull/636 was introduced, no builds have required the tools or Go versions this PR removes support for.

## Test migration

Several legacy tests covered general buildpack behavior (linker flags, `bin` file validation, `GOVERSION` override, `GO_INSTALL_PACKAGE_SPEC` override, `GO_INSTALL_TOOLS_IN_IMAGE`) that was only exercised through godep fixtures. These tests were migrated to use Go modules fixtures, preserving coverage for functionality that remains supported. In total, 64 legacy test functions were removed and one new test (`testDeprecatedToolDetected`) was added for the legacy tool migration error, bringing the suite from 99 to 36 tests.

## Changes

- `bin/detect`: Still detect legacy dependency manager files (for helpful compile-time errors), but only `go.mod` leads to a successful build
- `bin/compile`: Remove legacy tool functions, case branches, GOPATH setup, migration tool support, `concurrency.sh`, deprecation warnings, dead go1.0-go1.4 linker flag branch, and `LocalName`/file renaming logic. Add specific error messages for legacy tools with migration guidance.
- `lib/common.sh`: Remove legacy file variables, detection branches, env flags (`GO15VENDOREXPERIMENT`, `GLIDE_SKIP_INSTALL`, `GO_SETUP_GOPATH_IN_IMAGE`), `setGoVersionFromEnvironment`, `determinLocalFileName`, and pre-go1.11 version expansions. Add legacy tool detection in `determineTool` with `go_tool` build data tracking.
- `bin/test-compile`, `bin/test`: Modules-only code paths
- `data.json`: Remove legacy tool version entries, `NeedsConcurrency`, `SupportsVendorExperiment`, and pre-go1.11 version expansions
- `files.json`: Remove legacy tool entries, pre-go1.11 Go tarballs. Note that `jq-linux64` (unused since https://github.com/heroku/heroku-buildpack-go/pull/641), `stdlib.sh.v8` (unused since https://github.com/heroku/heroku-buildpack-go/pull/639) were also removed and added to new list of ignored files in the `make sync` target.
- `Makefile`: Update sync target to ignore removed S3 files
- `lib/common.sh`, `sbin/sync-files.sh`, `sbin/fetch-test-assets`: Simplify SHA validation to SHA256-only
- `.github/workflows/ci.yml`: Update container test fixtures
- Remove `vendor/concurrency.sh`, `LICENSE.godep`, legacy test functions and fixtures
- Update `README.md` (remove legacy sections, add `$GOVERSION` docs, update example) and `CHANGELOG.md`

## Future work

This PR paves the way for several improvements:

- **Simplify artifact hosting.** With the removal of nearly all non-Go artifacts from `files.json` (`golangci-lint-1.20.0-linux-amd64.tar.gz` remains as the only non-Go artifact currently in use), we should consider further simplifying the artifact download/hosting/syncing logic. In particular, we'll likely want to stop using our own S3 mirror and always download from the upstream source.
- **Consolidate tests.** The test suite is now 36 tests, down from 99. Of those 36, 12 are nearly identical `testModBasicGo1*` tests (go1.15 through go1.26) that each verify a specific Go release compiles a basic fixture - primarily serving to validate that a particular release is available in our mirror. These could be consolidated into 1–2 tests (e.g. one for the default version and one for an older supported version), freeing up ~10 tests worth of CI time. This becomes even more viable once we stop mirroring Go tarballs ourselves, since there would be no need to verify our mirror has each version.
- **Add arm64 support.** After migrating to upstream sources, adding arm64 artifacts to the inventory and aligning with similar implementations in our other buildpacks for inventory updates would be a natural next step.

GUS-W-18650766


bin/detect
```diff
@@ -7,13 +7,7 @@ buildpack=$(cd "$(dirname $0)/.." && pwd)
...
 source "${buildpack}/lib/common.sh"
 
-if test -f "${build}/go.mod" || #go modules
-   test -f "${build}/Gopkg.lock" || #dep
-   test -f "${build}/Godeps/Godeps.json" || # godeps
-   test -f "${build}/vendor/vendor.json" || # govendor
-   test -f "${build}/glide.yaml" || # glide
-   (test -d "${build}/src" && test -n "$(find "${build}/src" -mindepth 2 -type f -name '*.go' | sed 1q)") # gb
-then
+if test -f "${build}/go.mod"; then
```
edmorley: I think we might still want to pass detection here on the old package manager filenames, and instead handle the "we no longer support these Go package managers" error inside `bin/compile`?

The reason for this is that when an app doesn't have an explicit buildpack set (ie brand new app, or the user has run `heroku buildpacks:clear`) the build system runs the `bin/detect` of all default language buildpacks, and if none pass detection then [displays](https://github.com/heroku/cytokine/blob/85b200efb0ed5111028ff24e6c64fa9a99efbeef/cmd/cytokine/detect_buildpack.go#L37-L41) only the generic "can't detect the language error" - meaning the language-specific detection error messages aren't shown (since it's not really viable to show 8 of them, one from each language). That generic message directs people to the [docs here](https://devcenter.heroku.com/articles/buildpacks), and if they then figure out that how to explicitly add the `heroku/go` buildpack, only then will they finally see the stderr from `bin/detect`.

As such for Python I take the approach that `bin/detect` passing should be much more lenient (I match against 20+ different Python files), and instead only be checking whether the app is roughly a Python app (ie detecting just the intended language of the codebase, not whether it's a fully formed Python app). And then during `bin/compile` itself is where the "is this Python app correctly configured" checks run.

I then have two similar, but subtly different error message wordings:
1. [During `bin/detect`](https://github.com/heroku/heroku-buildpack-python/blob/96e5bea5603b78b40d6c2b59e738cc2b934d8d3c/bin/detect#L63-L97): ~your app is configured to use the Python buildpack but we can't find any Python-related files at all (since we know people will only be seeing the error message if they've explicitly configured the buildpack, and not in the generic default detection case)
2. [During `bin/compile`](https://github.com/heroku/heroku-buildpack-python/blob/96e5bea5603b78b40d6c2b59e738cc2b934d8d3c/lib/package_manager.sh#L95-L125): ~couldn't find the specific required Python package manager files

These are then tested via:
https://github.com/heroku/heroku-buildpack-python/blob/main/spec/hatchet/detect_spec.rb
https://github.com/heroku/heroku-buildpack-python/blob/main/spec/hatchet/package_manager_spec.rb#L6-L50

edmorley:
And thank you so much for this PR - great cleanup! :-D

bin/detect
runesoerensen: Good point - initially I just dropped detection entirely since these tools have been deprecated for well over a year by Heroku (and 6+ years individually by each project), and `bin/report` data confirmed no recent builds required them.

That said, the bigger concern might be related to outdated sample apps and documentation. Several Dev Center articles still reference legacy tools. If someone follows those as-is, they'd just get the generic "can't detect language" message instead of something actionable.

So as is usually the case, probably best to follow the Python buildpack's lead here :) I've pushed https://github.com/heroku/heroku-buildpack-go/pull/644/changes/97969a7b70cf03b20c1d5459d4c3d1e5de5637b3 with updated `bin/detect` to also pass on legacy dependency manager files (i.e. reverting the changes previously introduced), and `bin/compile` now gives a specific "support for X has been removed" error naming the detected tool and pointing to the Go modules docs. Also sets `go_tool` in build data so we can track this failure mode better.

I added `testDeprecatedToolDetected` (with a `dep-deprecated` fixture) to the existing test suite to cover this (as Hatchet tests haven't been implemented yet in this buildpack yet).

edmorley: Yeah also the `.go` file scan for maxdepth 2 (which has now been added back too) also catches the case where the user has nested the Go project inside a subdirectory in the repo.

lib/common.sh
```diff
@@ -344,8 +344,38 @@ determineTool() {
...
+            err ""
+            err "For more details see:"
+            err "https://devcenter.heroku.com/articles/go-modules"
+        else
+            err "A go.mod file is required."
+            err ""
+            err "For help with using Go on Heroku, see:"
+            err "https://devcenter.heroku.com/articles/go-support"
+        fi
         exit 1
```
edmorley, line 379: (for the future) I guess we need failure/error reasons in this buildpack at some point too :-)

edmorley (approved):
🎉

lib/common.sh
runesoerensen: For sure! And fortunately we now have a very easy way to track those :)

This particular failure/error mode should currently be possible to infer without explicitly setting a `failure_reason` though (since `bin/report` will print a `go_tool` value != `gomodules` for any app that passes `bin/detect`, and I think that can only happen here). That doesn't change the fact that explicitly recording the failure reasons will be more robust, here and elsewhere :)

PR #649, opened 2026-03-08
Remove redundant integration tests

runesoerensen:
This PR removes redundant `testModBasicGo1*` integration tests (`go1.15`-`go1.24`) and their fixtures. These nearly identical tests (only differing in the Go version tested) were primarily validating that each Go release is available in our S3 mirror, which we [recently migrated away from](https://github.com/heroku/heroku-buildpack-go/pull/647).

This reduces the total number of integration tests from 36 to 27, reducing CI execution time by about 25%.

Going forward, we're assuming that these Go releases will continue to be available from the upstream sources tracked in `files.json` without verifying it on every CI run.

The remaining three version-specific tests cover the oldest supported Go version (go1.11) and the last two major Go releases (go1.25 and go1.26):
* The go1.11 test ensures that the oldest supported `go mod` still works with the buildpack's args/env vars.
* The go1.25 and go1.26 tests cover actively maintained Go releases that receive frequent minor patch updates, ensuring the automated inventory PRs adding new entries to `files.json` are tested in CI before merging.

GUS-W-21497374


test/run.sh
```diff
@@ -251,146 +251,6 @@ testModBasic() {
   assertInstalledFixtureBinary
 }
 
-testModBasicGo115() {
```
edmorley, line 254: Is it worth keeping one test for "oldest supported Go version" (probably the oldest version in files.json?), just to make sure eg old go mod still works with whatever args/env vars the buildpack is passing/setting?

edmorley (approved):

test/run.sh
runesoerensen: Yes I think that's a good idea! I've pushed a commit adding a test for go1.11 https://github.com/heroku/heroku-buildpack-go/pull/649/changes/d846c60174bd609db48f175be7820c9e7770dfe7.

As noted in the commit, it uses the heroku `goVersion` comment instead of the `go` directive. That code path is tested in other tests, but since this PR is intended to only _consolidate_, it seems preferable to keep coverage for the way the version was specified in 7 of the fixtures removed (we changed to using the `go` directive in the `mod-basic-go122` fixture and later).

PR #653, opened 2026-03-10
Adjust curl retry max-time settings to handle transient failures

runesoerensen:
We've observed intermittent 502 errors and slow transfers when downloading Go binaries from `dl.google.com`. With `--retry-max-time` set to the same value as `--max-time` (150s), a single slow retry that hits the per-request timeout exhausts the entire retry budget, leaving no room for subsequent attempts.

This PR increases `--retry-max-time` from 200s, which provides enough budget for 4 file slow downloads that each hit the decreased 30s `--max-time` limit, while still leaving room for a successful retry under normal conditions (~2-3s) after default exponential backoff.

GUS-W-21524382

### DIFF
--- a/lib/common.sh
+++ b/lib/common.sh
@@ -16,7 +16,7 @@ NC='\033[0m' # No Color
 # seconds relying on the build system timeout. Go tarballs are up to ~70 MB and typically download in a few
 # seconds on Heroku, so we set relatively low timeouts to reduce delays before retries.
 # We use --no-progress-meter rather than --silent so that retry status messages are printed.
-CURL="curl --no-progress-meter --location --fail --max-time 150 --retry-max-time 150 --retry 5 --retry-connrefused --connect-timeout 5"
+CURL="curl --no-progress-meter --location --fail --max-time 150 --retry-max-time 350 --retry 5 --retry-connrefused --connect-timeout 5"
 
 TOOL=""
 # Default to $SOURCE_VERSION environment variable: https://devcenter.heroku.com/articles/buildpack-api#bin-compile
--- a/lib/common.sh
+++ b/lib/common.sh
@@ -16,7 +16,7 @@ NC='\033[0m' # No Color
 # seconds relying on the build system timeout. Go tarballs are up to ~70 MB and typically download in a few
 # seconds on Heroku, so we set relatively low timeouts to reduce delays before retries.
 # We use --no-progress-meter rather than --silent so that retry status messages are printed.
-CURL="curl --no-progress-meter --location --fail --max-time 150 --retry-max-time 350 --retry 5 --retry-connrefused --connect-timeout 5"
+CURL="curl --no-progress-meter --location --fail --max-time 30 --retry-max-time 200 --retry 5 --retry-connrefused --connect-timeout 5"
 
 TOOL=""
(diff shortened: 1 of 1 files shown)


lib/common.sh
```diff
@@ -16,7 +16,7 @@ NC='\033[0m' # No Color
 # seconds relying on the build system timeout. Go tarballs are up to ~70 MB and typically download in a few
 # seconds on Heroku, so we set relatively low timeouts to reduce delays before retries.
 # We use --no-progress-meter rather than --silent so that retry status messages are printed.
-CURL="curl --no-progress-meter --location --fail --max-time 150 --retry-max-time 150 --retry 5 --retry-connrefused --connect-timeout 5"
+CURL="curl --no-progress-meter --location --fail --max-time 150 --retry-max-time 350 --retry 5 --retry-connrefused --connect-timeout 5"
```
edmorley: Would we be better having the max-time lower? Even if the Go mirror is sometimes flaky, I presume normally retrying will get a quick download on the next attempt most times? Having `max-time` set to 150s means any slow CDN server can block the build for that long?

runesoerensen: @edmorley Yeah I think that'd make sense -- and considering the behavior I'm seeing it'd be better retry the download more times (e.g. 4 attempts with 60s (or lower) limit perhaps, and have the retry-max-time be something like 300s to allow for retry backoff etc).. Does that sound reasonable or what would you suggest?

`go.dev` does seem unusualy flaky right now, so and I can see already that more than 2 attempts is likely necessary.

Either way should definitely also improve the error with an actual error message (though the curl retry output is marginally improved now) :)

And of course consider alternative hosting/caching strategies (or even reverting to our S3 mirror) if this persists/happens frequently.

runesoerensen: Alright I've tried decreasing the max-time to 30s, and lowering the retry-max-time to 200s.. Basically this should give us plenty enough time for 6 attempts total (where the last one completes in a couple seconds) in a fairly unlikely scenario (hitting the 30s timeout 5 times).

Also had to make a change to avoid breaking test assertions (redirecting the retry messages now printed to stderr to stdout. Let me know if you have a better solution for that or this seems fine? :)

bin/compile
```diff
@@ -139,7 +139,7 @@ ensureGo() {
                         local sha=$(echo ${goVersion} | cut -d - -f 2)  #assumes devel-<short sha> or devel-<full sha>
                         local url="https://github.com/golang/go/archive/$sha.tar.gz"
                         start "Downloading development Go version ${goVersion}"
-                            ${CURL} ${url} | tar zxf -
+                            ${CURL} ${url} 2>&1 | tar zxf -
```
edmorley: I think this will break tar extraction, given the stdout is piped into tar and will contain retry messages?

runesoerensen: Ah yeah that seems right... Maybe we can just drop it here then? Not sure if it's too much of a hack -- but I'm pretty sure we're not testing this path anyway...

edmorley: Agree. And longer term perhaps we can drop support for development versions entirely?

runesoerensen: Yep sounds like a plan!

edmorley (approved):

PR #659, opened 2026-03-16
Support arm64

### LINKED ISSUE #557: Support ARM64 when using Heroku-24 Docker images
Heroku itself currently runs on AMD64 CPUs, however, some users use our buildpacks locally on machines with ARM64 CPUs (such as M1/M2/M3 MacBooks) with the Heroku base images published to Docker Hub. 

As such there have been requests to support the ARM64 architecture, e.g.:
https://github.com/heroku/base-images/issues/194

Starting with Heroku-24, the base images published to Docker Hub are now multi-architecture (AMD64 + ARM64), and our preview [Cloud Native Buildpacks](https://github.com/heroku/buildpacks) support ARM64 when using Heroku-24.

However, until CNBs leave preview there will still be users using our classic buildpacks with our base images from Docker Hub, so it would be ideal if we could add ARM64 support to our classic buildpacks too. This will not only make the images faster to run locally, but also avoid breaking local development workflows if users update to Heroku-24 and miss the mention in the stack upgrade notes about using `--platform linux/amd64` to force the architecture back to AMD64.

For example, the buildpack should support:

1. `git clone https://github.com/heroku/go-getting-started && cd go-getting-started`
2. Add a `Dockerfile` with the below contents.
3. `docker build --tag arm-test --platform linux/arm64 .`
4. `docker run --rm -e PORT=5001 -p 5001:5001 arm-test`
5. `curl localhost:5001`

```Dockerfile
FROM heroku/heroku:24-build as build
ENV STACK=heroku-24
COPY --chown=heroku . /app
WORKDIR /app
# This is after the COPY line so buildpack updates are picked up.
RUN mkdir -p /tmp/buildpack /tmp/cache /tmp/env \
  && curl https://buildpack-registry.s3.amazonaws.com/buildpacks/heroku/go.tgz \
    | tar -xz -C /tmp/buildpack
RUN /tmp/buildpack/bin/compile /app /tmp/cache /tmp/env

FROM heroku/heroku:24
COPY --from=build --chown=heroku /app /app
ENV HOME=/app
WORKDIR /app
CMD ["bash", "-c", "for f in .profile.d/*; do source \"${f}\"; done && go-getting-started"]
```

Currently when I try the above, I get this error (which is expected, since support for ARM64 hasn't yet been added):

```
$ docker build --tag arm-test --platform linux/arm64 .
...
 > [build 5/5] RUN /tmp/buildpack/bin/compile /app /tmp/cache /tmp/env:                                                     
0.082 -----> Fetching jq... done                                                                                            
1.685 -----> Fetching stdlib.sh.v8... done                                                                                  
2.283 ----->                                                                                                                
2.283        Detected go modules via go.mod                                                                                 
2.283 -----> 
2.285        Detected Module Name: github.com/heroku/go-getting-started
2.285 -----> 
2.319 -----> New Go Version, clearing old cache
2.319 -----> Installing go1.20.14
2.373 -----> Fetching go1.20.14.linux-amd64.tar.gz... done
10.53 rosetta error: failed to open elf at /lib64/ld-linux-x86-64.so.2
10.53  -----> Determining packages to install
10.53 rosetta error: failed to open elf at /lib64/ld-linux-x86-64.so.2
```

cc @joshwlewis

runesoerensen:
This PR adds ARM64 (linux-arm64) support to the buildpack, ensuring the correct Go toolchain and tools for the target platform are downloaded.

This enables users running `heroku-24` base images on ARM64 hosts (e.g. Apple Silicon Macs) to build without specifying `--platform linux/amd64` (as described in #557) - the buildpack now detects the architecture and fetches the matching toolchain automatically.

It also significantly improves local development performance by avoiding amd64 emulation. Benchmarks on an M4 Max MacBook Pro:

| Command | amd64 (emulated) | arm64 (native) | Speedup |
|---------|-------------------|-----------------|---------|
| `make test` | ~6:20 | ~2:10 | ~3x |
| `make test-parallel` (8 jobs) | 1:40 | 0:53 | ~2x |
| `make run-ci` (mod-deps-with-tests) | 0:34 | 0:15 | ~2x |
| `make run` (mod-basic-go126) | 0:21 | 0:07 | ~3x |

### Changes

- Detect architecture in `bin/compile` using `dpkg --print-architecture`
- Replace hardcoded `linux-amd64` in tarball filename construction with `linux-${ARCH}` in `ensureGo()` and the `golangci-lint` download
- Update `sbin/add-version` to add both amd64 and arm64 entries to `files.json`
- Populate `files.json` with arm64 entries for all tracked Go versions and golangci-lint
- Drop hardcoded `--platform linux/amd64` Docker flag in `Makefile`

Closes #557 

GUS-W-21599964

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,8 +20,7 @@ COMPILE_FAILURE_EXIT_CODE ?= 1
 
 # Converts a stack name of `heroku-NN` to its build Docker image tag of `heroku/heroku:NN-build`.
 STACK_IMAGE_TAG := heroku/$(subst -,:,$(STACK))-build
-# TODO: Add buildpack support for arm64 and use the native architecture for improved test performance locally.
-DOCKER_FLAGS := --rm --platform linux/amd64 -v $(PWD):/src:ro
+DOCKER_FLAGS := --rm -v $(PWD):/src:ro
 
 test:
 	@echo "Running tests using: STACK=$(STACK) TEST=$(TEST)"
--- a/bin/compile
+++ b/bin/compile
@@ -26,6 +26,8 @@ compile_start_time=$(build_data::current_unix_realtime)
 # a build report that will be consumed by the build system for observability.
 build_data::setup
 
+ARCH=$(dpkg --print-architecture)
+
 snapshotBinBefore
 
 # Clean up old cache files if migrating from old cache structure.
--- a/bin/compile
+++ b/bin/compile
@@ -118,12 +118,12 @@ ensureGo() {
 				local bGoVersion
 				# shellcheck disable=SC2310
 				bGoVersion="$(expandVer "${DefaultGoVersion}")"
-				goFile="${bGoVersion}.linux-amd64.tar.gz"
+				goFile="${bGoVersion}.linux-${ARCH}.tar.gz"
 				goPath="${GO_CACHE_DIR}/${bGoVersion}/go"
 				txt="Installing bootstrap ${bGoVersion}"
 				;;
 			*)
-				goFile="${goVersion}.linux-amd64.tar.gz"
+				goFile="${goVersion}.linux-${ARCH}.tar.gz"
 				;;
 		esac
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## [Unreleased]
 
 * Remove redundant `--retry-max-time` curl option
+* Add ARM64 (linux-arm64) architecture support
 
 ## [v226] - 2026-03-16
 
--- a/bin/test-compile
+++ b/bin/test-compile
@@ -19,5 +19,5 @@ if [[ -f "${BUILD_DIR}/.golangci.yml" ]] || [[ -f "${BUILD_DIR}/.golangci.toml" 
 	output::step "/.golangci.{yml,toml,json} detected"
 	tmp="$(mktemp -d)"
 	mkdir -p "${BUILD_DIR}/.heroku/golangci/bin"
-	ensureFile "golangci-lint-1.20.0-linux-amd64.tar.gz" "${tmp}" "tar -C ${BUILD_DIR}/.heroku/golangci/bin --strip-components=1 -zxf"
+	ensureFile "golangci-lint-1.20.0-linux-${ARCH}.tar.gz" "${tmp}" "tar -C ${BUILD_DIR}/.heroku/golangci/bin --strip-components=1 -zxf"
 fi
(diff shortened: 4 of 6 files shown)


Makefile
```diff
@@ -20,11 +20,12 @@ COMPILE_FAILURE_EXIT_CODE ?= 1
 
 # Converts a stack name of `heroku-NN` to its build Docker image tag of `heroku/heroku:NN-build`.
 STACK_IMAGE_TAG := heroku/$(subst -,:,$(STACK))-build
-# TODO: Add buildpack support for arm64 and use the native architecture for improved test performance locally.
-DOCKER_FLAGS := --rm --platform linux/amd64 -v $(PWD):/src:ro
+# Use the native platform architecture for improved performance by default.
+PLATFORM ?= linux/$(shell dpkg --print-architecture 2>/dev/null || uname -m | sed 's/x86_64/amd64/;s/aarch64/arm64/')
```
edmorley: (optional) Do we need all this platforms handling? For Python it just works locally on ARM without it? 
https://github.com/heroku/heroku-buildpack-python/blob/main/Makefile

(Albeit with one caveat I guess - if not using containerd and there is an AMD64 image already downloaded locally, then iirc that gets used rather than arm, until it's purged)

bin/compile
```diff
@@ -26,6 +26,9 @@ compile_start_time=$(build_data::current_unix_realtime)
 # a build report that will be consumed by the build system for observability.
 build_data::setup
 
+ARCH=$(dpkg --print-architecture)
+build_data::set_string "arch" "${ARCH}"
```
edmorley: This value will always be `amd64` on Heroku and will never vary - do we need it? It's also something not specific to the buildpack - so perhaps an attribute that should be set by cytokine/... like `stack` is at present (if we decided there was a use for it?).

edmorley (approved):

Makefile
runesoerensen: Ah good point and yes that behavior is the caveat/downside when not using containerd - but it's not really a problem as the tests will run anyway, and users can just remove the non-native image if they like. It was helpful to overwrite the value while testing/benchmarking, but I think relying on the default as you suggest is preferable here -- I've rewritten the commits to just removed the hardcoded platform flag instead https://github.com/heroku/heroku-buildpack-go/pull/659/changes/17e72ac503c4797b7d9778295703e68cd8c672b9

bin/compile
runesoerensen: No we don't need it - removed!

PR #679, opened 2026-07-29
Use canonical go.dev download URLs for Go binaries

runesoerensen:
## Summary

- Switch all Go download URLs in `files.json` from `dl.google.com` to the canonical `go.dev` host (478 entries), matching the Go CNB inventory. Stored SHA256 checksums are unchanged — each `go.dev/dl/<file>` was verified to 302-redirect to the previous `dl.google.com` URL.
- Update `sbin/add-version` to emit `go.dev` URLs and source the checksum from the `go.dev` release JSON (`?mode=json`), since the per-file `<tarball>.sha256` endpoint is not served from `go.dev`. This prevents the next version bump from re-introducing `dl.google.com` URLs.
- The two `golangci-lint` (`github.com`) entries are left unchanged.

GUS-W-23623855

### DIFF
--- a/sbin/add-version
+++ b/sbin/add-version
@@ -2,7 +2,7 @@
 
 set -euo pipefail
 
-BASEURL="https://dl.google.com/go"
+BASEURL="https://go.dev/dl"
 V="$1"
 
 if [[ -z "${V}" ]]; then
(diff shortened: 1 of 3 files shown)


sbin/add-version
```diff
@@ -2,21 +2,29 @@
...
 
 if [[ -z "${V}" ]]; then
 	echo "usage: $0 <go version>" >&2
 	exit 1
 fi
 
+# Fetch the release metadata once. go.dev serves the sha256 for each file
+# inline in this JSON; the per-file `<tarball>.sha256` endpoint is not served
+# from go.dev (it redirects to the docs page), so we source checksums here.
+if ! releases="$(curl -s -f "https://go.dev/dl/?mode=json&include=all")"; then
```
edmorley, line 16: (something for a future PR, given it's pre-existing) The curl usages here suppress error messages and also don't retry/have timeouts etc (so failures would mean a 24 hour delay before the Go updates cron next runs and delay opening a PR etc).

edmorley (approved):

sbin/add-version
runesoerensen: For sure -- pushed follow-up PR to adopt the approach from `sbin/update-go-versions` https://github.com/heroku/heroku-buildpack-go/pull/681

PR #681, opened 2026-07-29
Add curl timeouts and retries to sbin/add-version metadata fetch

runesoerensen:
## Summary

- Follow-up to #679 addressing [this review comment](https://github.com/heroku/heroku-buildpack-go/pull/679#discussion_r3673795151): the go.dev metadata `curl` in `sbin/add-version` used `-s -f`, which silenced errors and had no timeouts or retries.
- Replace it with the same resilient invocation already used in `sbin/update-go-versions`: `--no-progress-meter` (instead of `-s`, so retry status messages stay visible) plus connection/overall timeouts and retries (`--connect-timeout`, `--max-time`, `--retry`, `--retry-max-time`, `--retry-connrefused`).
- Without this, a transient go.dev blip fails the run and delays Go updates until the next daily cron.

> Note: based on the `switch-inventory-to-go-dev` branch since the go.dev `curl` line only exists there; will auto-retarget to `main` once #679 merges.

GUS-W-23636866

### DIFF
--- a/sbin/add-version
+++ b/sbin/add-version
@@ -13,7 +13,19 @@ fi
 # Fetch the release metadata once. go.dev serves the sha256 for each file
 # inline in this JSON; the per-file `<tarball>.sha256` endpoint is not served
 # from go.dev (it redirects to the docs page), so we source checksums here.
-if ! releases="$(curl -s -f "https://go.dev/dl/?mode=json&include=all")"; then
+# The curl flags mirror sbin/update-go-versions: --no-progress-meter (rather
+# than -s) keeps retry status messages visible, and the timeouts/retries stop a
+# transient go.dev blip from failing the run and delaying updates until the next
+# daily cron.
+if ! releases="$(curl \
+	--no-progress-meter \
+	--fail \
+	--connect-timeout 3 \
+	--max-time 60 \
+	--retry 5 \
+	--retry-max-time 60 \
+	--retry-connrefused \
+	'https://go.dev/dl/?mode=json&include=all')"; then
 	echo "error: adding ${V}: couldn't fetch release metadata from go.dev" >&2
 	exit 1
 fi


sbin/add-version
```diff
@@ -13,7 +13,19 @@ fi
...
+# daily cron.
+if ! releases="$(curl \
+	--no-progress-meter \
+	--fail \
+	--connect-timeout 3 \
+	--max-time 60 \
+	--retry 5 \
+	--retry-max-time 60 \
+	--retry-connrefused \
+	'https://go.dev/dl/?mode=json&include=all')"; then
```
edmorley, line 28: Since we no longer support pre-release versions, should this be `include=stable` to match `sbin/update-go-versions`?

```suggestion
	'https://go.dev/dl/?mode=json&include=stable')"; then
```

edmorley (approved):

sbin/add-version
runesoerensen: @edmorley Uhm maybe, but my motivation for using `include=all` were:

* The stable feed only includes releases for the last two major major releases. Usually there's about a week between the last minor revision for a major release, and the major release that renders it unsupported (e.g. `go1.24.13` was released 2026-02-04 an no longer appeared in the stable feed after the `go1.26.0` release on 2026-02-10), but there have been cases where this gap was shorter. In practice we'd likely have updated the inventory before it's removed from the stable feed, but if not we still have the option to add it by manually passing the version (e.g. [using workflow dispatch](https://github.com/heroku/heroku-buildpack-go/blob/main/.github/workflows/update-go-versions.yml#L9-L11))
* I think I'd be a bit surprised if I called a script named `sbin/add-version go1.24.13` and got an error that no metadata could be found for that version (even if it exists, but wasn't included in the stable feed). Instead it's currently the caller's responsibility to only add what they want to add, and this will usually be `sbin/update-go-versions`'s auto-discovered stable releases.
* There's no way a pre-release would be added unless a maintainer explicitly requested it.

Could go with stable feed, but not sure I see the benefit in restricting/enforcing the stable-only policy in this particular script?

At any rate, I suspect we'll eventually move to a similar model as the CNB, and rewrite the entire inventory on every update rather than specifying Go versions (either manually or auto-discovered, which would render the lookup in this script unnecessary :)

edmorley: @runesoerensen Ah I was thinking "stable" meant "everything apart from pre-releases", but if that's not the case then feel free to ignore :-)

runesoerensen: Ah yes "stable" in the feed means the latest minor revision in the two newest major releases (that's also why we can get away with using that feed it to [detect new versions to add relatively effortlessly](https://github.com/heroku/heroku-buildpack-go/blob/main/sbin/update-go-versions#L26) :)

PR #684, opened 2026-08-08
Ignore `capability[]` and `state[]` in the git credential helper

### LINKED ISSUE #683: Git credential helper fails on `capability[]` attribute sent by git >= 2.46, breaking private repo fetches on heroku-26
The credential helper installed by `setGitCredHelper` (lib/common.sh) hard-fails on any attribute it doesn't recognize:

```
*)
    echo "Unsupported key: ${key}=${value}" >&2
    exit 1
;;
```

git 2.46 introduced credential-protocol capabilities and now sends `capability[]=authtype` to helpers (and `state[]` can follow). The heroku-26 stack ships a git new enough to do this, so every private module/repo fetch that relies on `GO_GIT_CRED__*` dies during `Determining packages to install`:

```
go: downloading github.com/<org>/<private-repo> v0.3.0
...
    Unsupported key: capability[]=authtype
    fatal: could not read Username for 'https://github.com': terminal prompts disabled
```

The git-credential documentation requires helpers to silently ignore unrecognized attributes, the same way this helper already tolerates `wwwauth[]`. Because the helper is installed unconditionally, this fails even for apps that set `GO_GIT_CRED__HTTPS__GITHUB__COM` correctly — there is currently no working private-repo path on heroku-26.

Reproduction: heroku-26 stack, any private Go module dependency, `GO_GIT_CRED__HTTPS__GITHUB__COM` set per the README.

Fix is a one-liner (PR attached): treat `capability[]` and `state[]` like `wwwauth[]`. We've verified it on a production app — with the patched helper, the same build configuration fetches the private module and deploys successfully.

patjakubik:
Fixes #683.

git 2.46+ sends `capability[]=authtype` (and optionally `state[]`) to credential helpers as part of the credential-protocol capabilities handshake. The helper's attribute parser hard-fails on unrecognized keys, which breaks every `GO_GIT_CRED__*`-authenticated private repo fetch on stacks with newer git (heroku-26):

```
Unsupported key: capability[]=authtype
fatal: could not read Username for 'https://github.com': terminal prompts disabled
```

The git-credential documentation requires helpers to ignore unrecognized attributes. This change treats `capability[]` and `state[]` the same as the already-tolerated `wwwauth[]`.

Tested by pointing an app on heroku-26 with a private Go module dependency at this branch: the previously failing build fetches the module and deploys successfully; without the patch the same configuration fails as above.

### DIFF
--- a/lib/common.sh
+++ b/lib/common.sh
@@ -226,9 +226,9 @@ setGitCredHelper() {
                         password="${value}"
                     ;;
                     wwwauth[]|capability[]|state[])
-                        # Informational fields (git 2.44+/2.46+); the
-                        # credential-helper protocol requires ignoring
-                        # unrecognized attributes.
+                        # Informational fields:
+                        # - wwwauth[] (git 2.41+)
+                        # - capability[] / state[] (git 2.46+)
                         :
                     ;;
                     *)
--- a/lib/common.sh
+++ b/lib/common.sh
@@ -225,7 +225,10 @@ setGitCredHelper() {
                     password)
                         password="${value}"
                     ;;
-                    wwwauth[])
+                    wwwauth[]|capability[]|state[])
+                        # Informational fields (git 2.44+/2.46+); the
+                        # credential-helper protocol requires ignoring
+                        # unrecognized attributes.
                         :
                     ;;
                     *)


lib/common.sh
```diff
@@ -225,7 +225,10 @@ setGitCredHelper() {
                     password)
                         password="${value}"
                     ;;
-                    wwwauth[])
+                    wwwauth[]|capability[]|state[])
+                        # Informational fields (git 2.44+/2.46+); the
+                        # credential-helper protocol requires ignoring
+                        # unrecognized attributes.
```
colincasey: Not sure what the inclusion of `2.44+` is supposed to mean here. If it's to indicate when each field became available, maybe the comment should be changed to the following:

```suggestion
                        # Informational fields:
                        # - wwwauth[] (git 2.41+)
                        # - capability[] / state[] (git 2.46+)
```

colincasey:
@patjakubik thanks for this PR! Could you add:
- A brief CHANGELOG entry describing the change (see https://github.com/heroku/heroku-buildpack-go/pull/570 for a similar PR to this).
- Unit or integration tests (if possible).

colincasey:
Closed in favor of #687

